PR #10, opened 2022-05-31
Introduce a module with types/abstractions for working with VMSAs.

tylerfanelli:
Signed-off-by: Tyler Fanelli <tfanelli@redhat.com>

### DIFF
--- a/src/vmsa/mod.rs
+++ b/src/vmsa/mod.rs
@@ -1,0 +1,495 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Types and abstractions regarding Virtual Machine Save Areas (VMSAs).
+
+#![allow(dead_code)]
+
+use super::*;
+
+use serde_big_array::BigArray;
+use std::fs;
+use std::io;
+
+const ATTR_G_SHIFT: usize = 23;
+const ATTR_B_SHIFT: usize = 22;
+const ATTR_L_SHIFT: usize = 21;
+const ATTR_AVL_SHIFT: usize = 20;
+const ATTR_P_SHIFT: usize = 15;
+const ATTR_DPL_SHIFT: usize = 13;
+const ATTR_S_SHIFT: usize = 12;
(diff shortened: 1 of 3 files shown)


crobinso:
LGTM, the additions too.

When making changes like this, it will be helpful in the future to have one commit that is just code movement from sevctl to sev, then additional commits on top that change things, so it's easier to find what changed.

crobinso (approved):

src/vmsa/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+    pub fn to_file(&self, filename: &str) -> VmsaRWResult {
+        let vmsa_buf = match bincode::serialize(self) {
+            Ok(v) => v,
+            Err(e) => return VmsaRWResult::serialization_err(e),
+        };
+
+        const SIZE: usize = std::mem::size_of::<Vmsa>();
+
+        // Pad to 4096 bytes
+        let buf: &mut [u8] = &mut [0; SIZE];
```
crobinso, line 482: Just saw your sevctl PR and that the test suite is busted. SIZE here should be 4096, that will fix it

src/vmsa/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        let vmsa = match bincode::deserialize(&data[..]) {
+            Ok(v) => v,
+            Err(e) => return VmsaRWResult::serialization_err(e),
+        };
+
+        VmsaRWResult::read_success(vmsa)
+    }
+
+    // Save binary VMSA content to the passed filename.
```
crobinso, line 471: You can drop this comment, the one after it covers things

crobinso (changes requested):

tylerfanelli:
> LGTM, the additions too.
> 

Great, I'll merge.

> When making changes like this, it will be helpful in the future to have one commit that is just code movement from sevctl to sev, then additional commits on top that change things, so it's easier to find what changed.

Noted, I'll make sure to separate the transferring and additions in the future.

PR #12, opened 2022-06-06
vmsa: Fix SIZE to refer to size of Vmsa struct

tylerfanelli:
Signed-off-by: Tyler Fanelli <tfanelli@redhat.com>


crobinso (approved):
yeah that looks right

PR #13, opened 2022-06-13
vmsa API tweaks

crobinso:
This tweaks the newly added vmsa API to be a bit more consistent with other `sev` code

* Use standard io::Result and io::Error
* Use codicon encode and decode methods for serializing

This also added a VMSA file length check on file input, to match the padding we do when writing to file. See individual commit messages for more details.

### DIFF
--- a/src/vmsa/mod.rs
+++ b/src/vmsa/mod.rs
@@ -305,36 +305,6 @@ pub struct Vmsa {
     x87_state_gpa: u64,
 }
 
-/// An enum that encompasses all valid return types of VMSA reading/writing and
-/// serialization/deserialization.
-pub enum VmsaRWResult {
-    /// An error occurred in reading/writing a VMSA struct.
-    IoErr(io::Error),
-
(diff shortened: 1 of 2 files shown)


src/vmsa/mod.rs
```diff
@@ -305,33 +306,19 @@ pub struct Vmsa {
     x87_state_gpa: u64,
 }
 
-/// An enum that encompasses all valid return types of VMSA reading/writing and
-/// serialization/deserialization.
-pub enum VmsaRWResult {
-    /// An error occurred in reading/writing a VMSA struct.
-    IoErr(io::Error),
+impl codicon::Decoder<()> for Vmsa {
+    type Error = std::io::Error;
```
tylerfanelli, line 310: I see, we're able to remove the use of `bincode::Error` by instead having the serialization/deserialization return an `io::Error` instead. This is much cleaner than the previous approach.

tylerfanelli (approved):

PR #19, opened 2022-08-05
session: allow verifying an external digest

slp:
In some cases, such as when the digest of each launch measurement is
pre-generated beforehand, it's useful to be able to verify an
externally generated digest instead of having to pass the payload data
through Session<Measuring>::update_data().

This change implements Session<Measuring>::verify_with_digest(), which
behaves in the same way as Session<Measuring>::verify() but verifying
a digest that is provided as an argument to the method.

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -204,6 +204,24 @@ impl Session<Measuring> {
 
         session.verify(&digest, build, msr)
     }
+
+    /// Verifies the session's measurement against the AMD SP's measurement
+    /// using an externally generated digest.
+    pub fn verify_with_digest(
+        self,
+        build: Build,
+        msr: launch::sev::Measurement,
+        digest: &[u8],
+    ) -> Result<Session<Verified>> {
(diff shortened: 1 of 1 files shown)


src/session/mod.rs
```diff
@@ -204,6 +204,24 @@ impl Session<Measuring> {
 
         session.verify(&digest, build, msr)
     }
+
+    /// Verifies the session's measurement against the AMD SP's measurement
+    /// using an externally generated digest.
+    pub fn verify_with_digest(
+        mut self,
```
tylerfanelli, line 211: Clippy indicates that this variable doesn't have to be a mutable reference. Can you remove the `mut`?

tylerfanelli (changes requested):
LGTM and I'm fine with this change.

src/session/mod.rs
slp: Fixed, thanks!

PR #23, opened 2022-08-23
Launch API trait cleanups

tylerfanelli:
This PR eases the `AsRawFd` trait requirements to the `sev::Launcher` and `snp::Launcher`. Rather than enforce these traits, we can just have callers pass two `RawFd`s (for the VM's file descriptor and `/dev/sev`'s file descriptor) and avoid all the extra boilerplate that comes along with ensuring these traits are satisfied.

I still want to fix another issue upcoming, which is the enforcement of a lifetime on the `launch::Start` struct.

### DIFF
--- a/src/launch/sev.rs
+++ b/src/launch/sev.rs
@@ -33,13 +33,6 @@ pub struct Launcher<T> {
     sev: RawFd,
 }
 
-impl<T> Launcher<T> {
-    /// Give access to the vm fd to create vCPUs or such.
-    pub fn as_mut_vmfd(&mut self) -> &mut RawFd {
-        &mut self.vm_fd
-    }
-}
-
 impl Launcher<New> {
     /// Begin the SEV launch process.
     pub fn new(kvm: RawFd, sev: RawFd) -> Result<Self> {
--- a/src/launch/sev.rs
+++ b/src/launch/sev.rs
@@ -12,7 +12,7 @@ use crate::*;
 
 use std::io::Result;
 use std::mem::MaybeUninit;
-use std::os::unix::io::AsRawFd;
+use std::os::unix::io::RawFd;
 
 use bitflags::bitflags;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


slp:
Even though I proposed dropping the `AsRawFd` traits myself, I now have some doubts about it. My main concern here is that RawFd doesn't have a guaranteed lifetime, so if the caller only intends to keep `Launcher` around after opening `/dev/sev` with a `File`, the underlying file descriptor may be closed when the `File` gets out of context.

I'll play a bit with this next Monday and provide further feedback.

tylerfanelli:
> Even though I proposed dropping the `AsRawFd` traits myself, I now have some doubts about it. My main concern here is that RawFd doesn't have a guaranteed lifetime, so if the caller only intends to keep `Launcher` around after opening `/dev/sev` with a `File`, the underlying file descriptor may be closed when the `File` gets out of context.
> 
> I'll play a bit with this next Monday and provide further feedback.

So, in that case -- the `File` supplied by `Firmware` could be owned by `Launcher` and used while launch measuring is in progress. The one thing we would have to assume is that the `vm_fd` would be open for the entire launch process (a fair assumption, I think).

slp:
> > Even though I proposed dropping the `AsRawFd` traits myself, I now have some doubts about it. My main concern here is that RawFd doesn't have a guaranteed lifetime, so if the caller only intends to keep `Launcher` around after opening `/dev/sev` with a `File`, the underlying file descriptor may be closed when the `File` gets out of context.
> > I'll play a bit with this next Monday and provide further feedback.
> 
> So, in that case -- the `File` supplied by `Firmware` could be owned by `Launcher` and used while launch measuring is in progress. The one thing we would have to assume is that the `vm_fd` would be open for the entire launch process (a fair assumption, I think).

Yes, keeping `AsRawFd` allows users to decide whether they want to give `Launcher` ownership of the object wrapping the file descriptor (i.e. `File`, which also provides the ability of getting clone via `try_clone()`, which translates on Linux to a `dup()`), or just go with a bare `RawFd` (i.e. `Launcher<RawFd, RawFd, Started>`).

tylerfanelli:
Closing this and noting that we'll keep the `AsRawFd` trait.

PR #28, opened 2022-09-06
launch/sev: Remove lifetime requirement

tylerfanelli:
Instead of a `sev::Launcher` taking references to `AsRawFd` implementers, allow then to take ownership if they see fit. If a caller does not want a `sev::Launcher` to take ownership, they can use `as_raw_fd()`.


tylerfanelli:
cc @slp

slp:
LGTM.

slp (approved):

PR #31, opened 2022-10-13
README: Fix links

sameo:
The links are missing an `src` prefix.

Signed-off-by: Samuel Ortiz <sameo@rivosinc.com>


tylerfanelli:
Thanks for pointing this out, @sameo.

Could you also change the links in the top comments of `lib.rs` and then run `cargo readme`? This will fix the `lint / cargo readme` failure.

PR #32, opened 2022-10-18
snp: Adding extended report features

larrydewey:
- Added in FFI wrappers for Snp Extended Report Features.
- Added in user apis for interacting with the FFI.
- Added in some unit tests and system tests for validating behavior.

### DIFF
--- a/src/firmware/linux/host/ioctl.rs
+++ b/src/firmware/linux/host/ioctl.rs
@@ -14,16 +14,16 @@ use std::marker::PhantomData;
 // source code: include/uapi/linux/psp-sev.h
 impl_const_id! {
     pub Id => u32;
-    PlatformReset = 0,
-    PlatformStatus = 1,
-    PekGen = 2,
-    PekCsr<'_> = 3,
-    PdhGen = 4,
-    PdhCertExport<'_> = 5,
-    PekCertImport<'_> = 6,
-    GetId<'_> = 8, /* GET_ID2 is 8, the deprecated GET_ID ioctl is 7 */
+    PlatformReset = 0x0,
+    PlatformStatus = 0x1,
+    PekGen = 0x2,
+    PekCsr<'_> = 0x3,
+    PdhGen = 0x4,
+    PdhCertExport<'_> = 0x5,
+    PekCertImport<'_> = 0x6,
+    GetId<'_> = 0x8, /* GET_ID2 is 0x8, the deprecated GET_ID ioctl is 0x7 */
 
-    SnpPlatformStatus = 9,
+    SnpPlatformStatus = 0x9,
     SnpSetExtConfig = 0xA,
     SnpGetExtConfig = 0xB,
 }
--- a/src/firmware/linux/host/ioctl.rs
+++ b/src/firmware/linux/host/ioctl.rs
@@ -24,6 +24,8 @@ impl_const_id! {
     GetId<'_> = 8, /* GET_ID2 is 8, the deprecated GET_ID ioctl is 7 */
 
     SnpPlatformStatus = 9,
+    SnpSetExtConfig = 0xA,
+    SnpGetExtConfig = 0xB,
 }
 
 const SEV: Group = Group::new(b'S');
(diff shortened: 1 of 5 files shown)


tylerfanelli:
Thanks @larrydewey, I'll give a review.

src/firmware/linux/host/types.rs
```diff
@@ -250,3 +255,218 @@ impl SnpConfig {
...
+                0
+            } else {
+                &CertTable::from_uapi(data.certs.clone().unwrap()) as *const CertTable as u64
+            },
+            config_address: if data.config.is_none() {
+                0
+            } else {
+                &data.config.unwrap() as *const SnpConfig as u64
+            },
+            certs_len: data.certs_len,
```
tylerfanelli, line 426: Do you think it's necessary to add a check, if `certs_address.is_none()` then `certs_len` should be 0?

larrydewey: I changed it to:

```rust
certs_len: if data.certs.is_none() {
    0
} else {
    data.certs_len
},
```

tylerfanelli:
Thanks! LGTM

src/firmware/linux/host/ioctl.rs
```diff
@@ -24,6 +24,8 @@ impl_const_id! {
     GetId<'_> = 8, /* GET_ID2 is 8, the deprecated GET_ID ioctl is 7 */
 
     SnpPlatformStatus = 9,
+    SnpSetExtConfig = 0xA,
+    SnpGetExtConfig = 0xB,
```
slp, line 28: Nitpick: For consistency, I think either those new values should be specified as decimal, or the already present ones moved to hexadecimal.

src/firmware/linux/host/types.rs
```diff
@@ -250,3 +255,222 @@ impl SnpConfig {
...
+}
+
+impl CertTable {
+    pub fn from_uapi(table: UapiCertTable) -> Self {
+        let mut entries: Vec<CertTableEntry> = vec![];
+        let mut tmp_guid: [u8; 16] = [0u8; 16];
+
+        for entry in table.entries {
+            // We have a problem if the GUID is anything other than 16 bytes.
+            assert_eq!(entry.guid().len(), 16);
```
slp, line 307: Given that `table` is a value provided as an argument, wouldn't it be better to change this function to return a `Result` and fail gracefully instead of using `assert`?

tylerfanelli:
It also seems that `cargo clippy` is failing now (which is new). Can you address this?

src/firmware/linux/host/ioctl.rs
larrydewey: I switched them all to hex for consistency.

src/firmware/linux/host/types.rs
larrydewey: I added in some error handling for this. Let me know what you think @slp.

src/firmware/uapi/host/types.rs
```diff
@@ -25,6 +25,76 @@ pub enum Indeterminate<T: Debug> {
...
+    }
+}
+
+impl std::fmt::Display for UserApiError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let err_msg: String = match self {
+            UserApiError::FirmwareError(error) => format!("Firmware Error Encountered: {}", error),
+            UserApiError::ApiError(error) => format!("Certificate Error Encountered: {}", error),
+        };
+        write!(f, "{err_msg}")
```
tylerfanelli, line 53: I think this should be something like

`write!(f, "{}", err_msg)`

larrydewey: The way I have it should be okay. Identifiers in strings were introduced in 1.58.0 ([Announcing Rust 1.58.0 - Identifiers in Strings](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings)).

tylerfanelli: Ah I see, my mistake. In that case, LGTM!

Issue #43, opened 2022-11-14
Add helper methods to the SnpExtReportReq type


larrydewey:
In hind-sight it would have been really good to add in some helper methods to get a more Rust friendly access to the data located at the pointers returned from the kernel here. I will create a PR to add some of this functionality in.

- [x] A obtain a copy of the certificates once the snp_get_ext_report() call had been completed (vector of CertTableEntry, etc).

dubek:
Another way to look at this: can we provide a rust-friendly `snp_get_ext_report()`, something like:

```
fn snp_get_ext_report(report_data: [u8; 64]) -> (AttestationReport, Certs)
```

It would automatically allocate a buffer for the certificates (maybe ioctl once with 0 to get the expected length and then ioctl again after allocating enough room). I basically want to hide all the low-level C stuff (pointers, buffer lengths).

Another thing to consider: in some places (for example, confidential containers KBC), the guest fetches the attestation report and certs but doesn't parse them; it just passes them as opaque buffers to the attestation-server/verifier (which then needs to parse them). In that usage it makes sense to have a version that just returns `Vec<u8>`s (or make the AttestationReport and CertTableEntry serializable back to bytes, but that feels like "double work").

(@fitzthum FYI)

fitzthum:
Yeah I think that signature is exactly what we want. Might add a VMPL level as input. It would be nice to avoid having to make a nested report request and extended report request.

fitzthum:
My impression is that certificate parsing will usually be handled remotely by the verifier. The guest really just needs to send the certs off somewhere else. So probably we should just return certs as a `Vec<u8>`.

tylerfanelli:
@fitzthum I see, so hiding away the low-level C stuff (pointers, buffer lengths) would only be done for the regular `snp_get_ext_report()` returning `(AttestationReport, Certs)`. Not for an API returning `(Vec<u8>, Vec<u8>)`.

fitzthum:
We want to hide the pointers no matter what. I'm just saying that we probably don't need to worry about parsing the certs. Currently we get the certs by making an extended report request and then unsafely copying from the provided buffer (with length). Instead it would be nice to just get a Vec<u8> with the same data and never have to touch the pointer.

larrydewey:
@fitzthum I agree, hiding the pointers is definitely important. I just missed this last part in my former PR. I decided to change the return type to:

```rust
-> Result<(AttestationReport, CertTable), Indeterminate<Error>> {
```

[...]

PR #45, opened 2022-11-15
Update Extended Report APIs

larrydewey:
This is a set of commits to update the extended report API to return the contents of the report and the certificates in a rust-friendly way. It also adds derived serialization and de-serialization to the associated data types to allow easier marshaling between services.

Addresses #43 

### DIFF
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -90,15 +90,7 @@ impl Firmware {
 
         SNP_GET_EXT_REPORT.ioctl(
             &mut self.0,
-            &mut SnpGuestRequest::new(
-                message_version,
-                &SnpExtReportReq {
-                    data: ext_report_request.data,
-                    certs_address: ext_report_request.certs_address,
-                    certs_len: ext_report_request.certs_len,
-                },
-                &ext_report_response,
-            ),
+            &mut SnpGuestRequest::new(message_version, &ext_report_request, &ext_report_response),
         )?;
 
         Ok(ext_report_response)
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -80,13 +80,15 @@ impl Firmware {
     /// # Arguments
     ///
     /// * `message_version` - (Optional) Used for the SnpGuestRequest, specifies the message version number defaults to 1.
-    /// * `ext_report_request` - an SnpExtReportReq object with its associated data for requesting the extended attestation report.
+    /// * `report_request` - an SnpReportReq object with its associated data.
     ///
     pub fn snp_get_ext_report(
         &mut self,
         message_version: Option<u8>,
-        ext_report_request: SnpExtReportReq,
+        report_request: SnpReportReq,
     ) -> Result<(AttestationReport, CertTable), Indeterminate<Error>> {
+        let ext_report_request: SnpExtReportReq = SnpExtReportReq::new(report_request);
+
         let ext_report_response: SnpReportRsp = SnpReportRsp::default();
         let mut cert_table: CertTable = Default::default();
 
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -9,7 +9,8 @@ use std::fs::{File, OpenOptions};
 
 use self::types::*;
 
-use super::host::types::{Error, Indeterminate};
+use super::host::types::{CertTable, Error, Indeterminate};
+use super::linux::host::types::CertTable as FFICertTable;
 use crate::firmware::linux::guest::ioctl::*;
 
 /// A handle to the SEV, SEV-ES, or SEV-SNP platform.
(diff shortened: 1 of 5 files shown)


fitzthum (approved):
Nice. Thanks @larrydewey  

I think there are still some tweaks that can be made to the inputs of the report functions, but this is handy for now.

src/firmware/guest/mod.rs
```diff
@@ -85,22 +86,21 @@ impl Firmware {
         &mut self,
         message_version: Option<u8>,
         ext_report_request: SnpExtReportReq,
```
dubek, line 88: with this, the end-user still has to fill `certs_address` and `certs_len`, which is a bit C-like which we'll prefer to hide.

dubek: we might want to get only the data field from the user; we can allocate a certs buffer inside the function.

tylerfanelli: Right, really only looking for a list of certs.

src/firmware/guest/mod.rs
```diff
@@ -85,22 +86,21 @@ impl Firmware {
...
-    ) -> Result<SnpReportRsp, Indeterminate<Error>> {
+    ) -> Result<(AttestationReport, CertTable), Indeterminate<Error>> {
```
tylerfanelli, line 89: Eventually, A `Vec<UapiCertTableEntry>` is returned. But the `CertTableEntry` includes offset + pointer field, which is something we'd like to avoid presenting to a user (as Dov noted). We'd instead like to build a `Vec` of the Certificates themselves, and not worry about the other data in `CertTableEntry`.

src/firmware/guest/mod.rs
```diff
@@ -85,22 +86,21 @@ impl Firmware {
...
-                &ext_report_response,
-            ),
+            &mut SnpGuestRequest::new(message_version, &ext_report_request, &ext_report_response),
         )?;
 
-        Ok(ext_report_response)
+        if let Some(linux_cert_table) =
+            unsafe { (ext_report_request.certs_address as *const FFICertTable).as_ref() }
+        {
+            cert_table = linux_cert_table.to_uapi();
```
tylerfanelli, line 101: Here is where a `UapiCertTableEntry` will be returned, but some other `Vec<Certificate>` should be returned from the `CertTableEntry` entries.

tylerfanelli (changes requested):
Like mentioned before, there's still some of the low-level kernel details of getting a report that a user doesn't care much for.

src/firmware/guest/mod.rs
larrydewey: @tylerfanelli I think there is some confusion about which data type is being used here. The `CertTableEntry` in question is the definition provided in the UAPI level. It looks like this:

```rust 
/// Certificates which are accepted for [`CertTableEntry`]
pub enum SnpCertType {
    /// AMD Root Signing Key (ARK) certificate
    ARK,

    /// AMD SEV Signing Key (ASK) certificate
    ASK,

    /// Versioned Chip Endorsement Key (VCEK) certificate
    VCEK,

    /// Other (Specify GUID)
    OTHER(String),
}

/// An entry with information regarding a specific certificate.
pub struct CertTableEntry {
    /// A Specificy certificate type.
    cert_type: SnpCertType,

    /// The raw data of the certificate.
    data: Vec<u8>,
}

/// Certificates to send to the PSP.
pub struct CertTable {
    /// A vector of [`CertTableEntry`].
    pub entries: Vec<CertTableEntry>,
}
```

There is no offset or pointer field returned, and the `entries` field of the `CertTable` is the `Vec<CertTableEntry>` shown above.

src/firmware/guest/mod.rs
larrydewey: If you review my comment above, the purpose of the `to_uapi()` function is to parse the certificates into the Rust-friendly structure provided.

larrydewey:
> Like mentioned above, there's still some of the low-level syscall details of getting a report that a user doesn't care much for.

Again, I am not entirely sure what low-level syscall details you are referring to?

src/firmware/guest/mod.rs
larrydewey: Yeah, I can definitely push a commit to make this change.

src/firmware/guest/mod.rs
tylerfanelli: I'm sorry, you are right. I was mistakenly looking in `linux/host/types.rs`, my mistake.

tylerfanelli (approved):

src/firmware/guest/mod.rs
```diff
@@ -79,28 +80,29 @@ impl Firmware {
...
+    /// * `report_request` - an SnpReportReq object with its associated data.
     ///
     pub fn snp_get_ext_report(
         &mut self,
         message_version: Option<u8>,
-        ext_report_request: SnpExtReportReq,
-    ) -> Result<SnpReportRsp, Indeterminate<Error>> {
+        report_request: SnpReportReq,
+    ) -> Result<(AttestationReport, CertTable), Indeterminate<Error>> {
+        let ext_report_request: SnpExtReportReq = SnpExtReportReq::new(report_request);
```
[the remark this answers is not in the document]
larrydewey: @dubek very good point. Like we talked about during the meeting, I misunderstood the behavior of the IOCTL. I am working on a patch to fix this issue, and will tag you on it when it is ready for review.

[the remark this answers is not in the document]
slp: > Related: We should also consider how we automatically test this. It will include spinning an SNP guest and running the test code in it. Do we have any idea how to approach this?

We could easily use libkrun for running some tests in a CI pipeline. The biggest problem is, as usual, the availability of publicly-accessible SNP-capable systems to run such pipeline.

PR #47, opened 2022-11-16
cargo: Change some dependencies

tylerfanelli:


crobinso (approved):
In the future, a bit of explanation in the commit message would be good IMO

tylerfanelli:
@crobinso I agree, I left some comments on the commit messages. Can you re-review?

Cargo.toml
```diff
@@ -35,8 +35,8 @@ dirs = "4.0"
 iocuddle = "0.1"
 serde-big-array = "0.4.1"
 kvm-ioctls = "0.11"
-static_assert_macro = "^1.1"
-bitfield = "^0.14"
+static_assertions = "^1.1.0"
+bitfield = "^0.13"
```
larrydewey, line 39: Is there any particular reason we are downgrading this from `0.14.x` to `0.13.x`?

crobinso (approved):

larrydewey (commented):
Other than my question about `biftield`, it LGTM.

Cargo.toml
tylerfanelli: We don't use anything from the `0.14` release, and reverting to `0.13` would make things simpler in downstream releases.

PR #51, opened 2022-12-01
policy: Fixing bitfield bugs

larrydewey:
Previously there was no way to access the policy or platform info fields. This patches those fields.

### DIFF
--- a/src/firmware/linux/guest/types.rs
+++ b/src/firmware/linux/guest/types.rs
@@ -321,22 +321,30 @@ bitfield! {
     /// |--------|---------------|----------------------------------------------------------------------------------------------------------------|
     /// | 7:0    | ABI_MINOR     | The minimum ABI minor version required for this guest to run.                                                  |
     /// | 15:8   | ABI_MAJOR     | The minimum ABI major version required for this guest to run.                                                  |
-    /// | 16     | SMT           | 0: SMT is disallowed.<br>1: SMT is allowed.                                                               |
+    /// | 16     | SMT           | 0: Host SMT usage is disallowed.<br>1: Host SMT usage is allowed.                                                               |
     /// | 17     | -             | Reserved. Must be one.                                                                                         |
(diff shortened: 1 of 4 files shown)


src/firmware/linux/guest/types.rs
```diff
@@ -331,12 +331,19 @@ bitfield! {
...
-    get_migrate_ma, set_migrate_ma: 18, 18;
-    get_debug, set_debug: 19, 19;
-    get_single_socket, set_single_socket: 20, 20;
+
+    /// The API_MINOR value contained in the 64 bit value.
+    pub api_minor, _: 7, 0;
+    /// The API_MAJOR value contained in the 64 bit value.
+    pub abi_major, _: 15, 8;
+    /// The SMT value contained in the 64 bit value.
+    pub smt, _: 16, 16;
```
fitzthum, line 340: Would it be clearer if this were `smt_allowed`?

fitzthum (approved):
LGTM

tylerfanelli (changes requested):
Although the actual code LGTM, do you think the comments should change? Instead of having something like:
`/// The API_MINOR value contained in the 64 bit value`

I think that the comment should explain what that value means, for example:
`/// API_MINOR field: Indicates the minor API version`

With this, someone reading the source code can have an idea on what each value indicates without having to look it up in the SNP spec.

What do you think?

larrydewey:
> Although the actual code LGTM, do you think the comments should change? Instead of having something like: `/// The API_MINOR value contained in the 64 bit value`
> 
> I think that the comment should explain what that value means, for example: `/// API_MINOR field: Indicates the minor API version`
> 
> With this, someone reading the source code can have an idea on what each value indicates without having to look it up in the SNP spec.
> 
> What do you think?

Yeah, that is a pretty easy set of changes. I have applied it in the latest commit. Let me know what you think.

tylerfanelli (approved):
Thanks!

PR #54, opened 2023-01-18
43 patch ext report buffer

larrydewey:
Adding in final changes for the extended report. These were tested on a milan box running 1.51.3. Final thing to test is the guest extended report call.


larrydewey:
Now that the [SEV Secure Nested Paging
Firmware ABI Specification version 1.54 ](https://www.amd.com/system/files/TechDocs/56860.pdf) has been published, and the [SEV-SNP 1.54 Firmware](https://developer.amd.com/wp-content/resources/amd_sev_fam19h_model0xh_1.54.01.zip) has been published, I will also make sure everything works as-is before removing the WIP, as well.

larrydewey:
> Now that the [SEV Secure Nested Paging Firmware ABI Specification version 1.54 ](https://www.amd.com/system/files/TechDocs/56860.pdf) has been published, and the [SEV-SNP 1.54 Firmware](https://developer.amd.com/wp-content/resources/amd_sev_fam19h_model0xh_1.54.01.zip) has been published, I will also make sure everything works as-is before removing the WIP, as well.

This has been verified working in the latest firmware.

src/firmware/linux/host/types.rs
```diff
@@ -285,61 +284,83 @@ pub struct CertTableEntry {
...
+    /// offset (1)  --> | RawCertificate_1 |
+    ///                 | ...              |
+    ///                 | ...              |
+    /// offset (2)  --> | RawCertificate_2 |
+    ///                 | ...              |
+    ///                 | ...              |
+    /// offset (n)  --> | RawCertificate_n |
+    ///                 |------------------|
+    ///          
+    /// ```
```
dubek, line 312: nice!

src/firmware/linux/host/types.rs
```diff
@@ -356,54 +377,59 @@ impl CertTable {
...
-        let mut cert_end: *mut u8;
+
+        // Start parsing the PSP data from the pointers.
+        let mut entry: CertTableEntry;
 
         loop {
-            let mut cert_bytes: Vec<u8> = vec![];
+            // Dereference the pointer to parse the table data.
+            entry = *data;
+            guid = Uuid::from_slice(entry.guid.as_slice()).unwrap();
```
dubek, line 400: Consider `?` instead of `.unwrap()` -- not sure that a parsing problem should panic the entire program. This change will require returning a `Result<...>`.

src/firmware/linux/host/types.rs
```diff
@@ -285,61 +284,83 @@ pub struct CertTableEntry {
...
+    ///                 | ...              |
+    /// offset (2)  --> | RawCertificate_2 |
+    ///                 | ...              |
+    ///                 | ...              |
+    /// offset (n)  --> | RawCertificate_n |
+    ///                 |------------------|
+    ///          
+    /// ```
+    ///
+    pub fn uapi_to_vec_bytes(table: &mut Vec<UapiCertTableEntry>) -> Result<Vec<u8>, SnpCertError> {
```
dubek, line 314: It would be helpful to have a unit-test for this function and for `parse_table`. They do not rely on any kernel-y/ioctl-y stuff, so should be a bit easier to supply some cert table and convert it to bytes and verify that result, and the other way around for parsing.
(sorry if I missed existing tests somewhere)

dubek (commented):
Impressive work building and parsing those structures. See some comments below.

Regarding the additional changes in SNP spec 1.54: the changes to existing commands are backward compatible, so if we keep the reserved fields as 0, we'll get the default behaviour (which is use VCEK, unless VLEK is set up). Since some changes also require support for host kernel, they can wait. So IMO firmware 1.54 changes should not be in this PR.

larrydewey:
@dubek Thanks for the review! After I finish making the final changes for the guest side of retrieving the extended guest request, I will get started addressing them!

src/firmware/linux/host/types.rs
larrydewey: That is definitely on my todo list. I want to test each of these functions as much as possible; __especially__ where there is no hardware or kernel requirements.

larrydewey:
> Impressive work building and parsing those structures. See some comments below.
> 
> Regarding the additional changes in SNP spec 1.54: the changes to existing commands are backward compatible, so if we keep the reserved fields as 0, we'll get the default behaviour (which is use VCEK, unless VLEK is set up). Since some changes also require support for host kernel, they can wait. So IMO firmware 1.54 changes should not be in this PR.

I agree, I have no intentions of including the VLEK support in this PR. I was merely testing that the functionality still working as expected after the firmware upgrade.

src/firmware/host/types.rs
```diff
@@ -492,32 +505,50 @@ pub struct SnpExtConfig {
...
-    pub certs: Option<CertTable>,
+    pub certs: Option<Vec<CertTableEntry>>,
 
     /// SET:
     ///     Length of the certificates.
     ///
     /// GET:
     ///     Length of the buffer which will hold the fetched certificates.
-    pub certs_len: u32,
+    pub certs_buf: u32,
```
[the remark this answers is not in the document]
larrydewey: My issue with using `certs_len`, is that it holds two meanings. First, the length of the buffer being pointer to by the `certs_address` field (the correct one). Second, the number of certificates being supplied. I am open to keeping it as certs_len, though.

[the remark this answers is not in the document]
dubek: The [documentation for the SNP_GET_EXT_REPORT ioctl](https://github.com/torvalds/linux/blob/master/Documentation/virt/coco/sev-guest.rst#23-snp_get_ext_report) says:

> On success, the snp_ext_report_resp.data will contain the attestation report and snp_ext_report_req.certs_address will contain the certificate blob. If the length of the blob is smaller than expected then snp_ext_report_req.certs_len will be updated with the expected value.

So I think that the `certs_len` field always holds length in bytes (never number of certs). When calling it holds the number of bytes in the certs_addresses buffer, and if that's not enough, on output it will hold how many bytes are needed.

Similarly for `SNP_GET_EXT_CONFIG` (hypervisor command).

So I'd go with `certs_len` -- which also matches the names exposed in the [kernel headers](https://github.com/torvalds/linux/blob/7287904c8771b77b9504f53623bb477065c19a58/include/uapi/linux/sev-guest.h#L59-L67):

```c
struct snp_ext_report_req {
	struct snp_report_req data;

	/* where to copy the certificate blob */
	__u64 certs_address;

	/* length of the certificate blob */
	__u32 certs_len;
};
```

src/firmware/guest/mod.rs
```diff
@@ -86,23 +86,62 @@ impl Firmware {
         &mut self,
         message_version: Option<u8>,
         report_request: SnpReportReq,
-    ) -> Result<(AttestationReport, CertTable), Indeterminate<Error>> {
+    ) -> Result<(AttestationReport, Vec<CertTableEntry>), Indeterminate<Error>> {
+        // Due to the complex buffer allocation, we will take the SnpReportReq
+        // provided by the caller, and create an extended report request object
+        // for them.
         let ext_report_request: SnpExtReportReq = SnpExtReportReq::new(report_request);
```
dubek, line 93: 1. This creates an SnpExtReportReq in which both certs_address and certs_len are zero. This will cause the sev-guest ioctl to not fill in certificates information at all (see [sev-guest.c](https://github.com/torvalds/linux/blob/7287904c8771b77b9504f53623bb477065c19a58/drivers/virt/coco/sev-guest/sev-guest.c#L504-L506) in the kernel).

   In AMD's sev-guest repo in get-report.c they set `certs_address = (u64)-1` and `certs_len = 0` to signal to the kernel module that the caller wants to know the needed length.

   Maybe modify `SnpExtReportReq::new` to put `u64::MAX` in `certs_address` ?

2. Should this variable be `mut` because the ioctl might modify its content?

src/firmware/guest/mod.rs
```diff
@@ -86,23 +86,62 @@ impl Firmware {
...
 
-        let ext_report_response: SnpReportRsp = SnpReportRsp::default();
-        let mut cert_table: CertTable = Default::default();
-
-        SNP_GET_EXT_REPORT.ioctl(
-            &mut self.0,
-            &mut SnpGuestRequest::new(message_version, &ext_report_request, &ext_report_response),
-        )?;
+        // Create an object for the PSP to store the response content in.
+        let ext_report_response: SnpReportRsp = Default::default();
```
dubek, line 96: Should this variable be `mut` because the ioctl will modify its content?

src/firmware/guest/mod.rs
tylerfanelli: > 1. This creates an SnpExtReportReq in which both certs_address and certs_len are zero. This will cause the sev-guest ioctl to not fill in certificates information at all (see [sev-guest.c](https://github.com/torvalds/linux/blob/7287904c8771b77b9504f53623bb477065c19a58/drivers/virt/coco/sev-guest/sev-guest.c#L504-L506) in the kernel).
>    In AMD's sev-guest repo in get-report.c they set `certs_address = (u64)-1` and `certs_len = 0` to signal to the kernel module that the caller wants to know the needed length.
>    Maybe modify `SnpExtReportReq::new` to put `u64::MAX` in `certs_address` ?

Yes, I think this would work well.

> 2. Should this variable be `mut` because the ioctl might modify its content?

I believe it should.

src/firmware/guest/mod.rs
tylerfanelli: I suppose it wouldn't hurt, but I'm not entirely sure it's needed. When we use this response to build an `SnpGuestRequest` [here](https://github.com/virtee/sev/pull/54/files#diff-f2a75c2a321a96e5bda859cce9503572e03a5068d455ee0c5a276dad6d949800R105), the request takes a pointer to this struct. The ioctl itself is inherently unsafe, as Rust can't guarantee any safety measures with what the kernel does with this data.

src/firmware/guest/mod.rs
```diff
@@ -86,23 +86,62 @@ impl Firmware {
...
+
+        let mut certificates: Vec<CertTableEntry> = vec![];
+
+        // Construct the needed object needed to perform the IOCTL request.
```
tylerfanelli, line 100: s/needed object needed/object needed

src/firmware/host/types.rs
```diff
@@ -79,6 +80,12 @@ pub enum SnpCertError {
     /// Malformed GUID.
     InvalidGUID,
 
+    /// Malformed Page Allignment
+    PageMisallignment,
```
tylerfanelli, line 84: s/PageMisallignment/PageMisalignment

src/firmware/host/types.rs
```diff
@@ -79,6 +80,12 @@ pub enum SnpCertError {
     /// Malformed GUID.
     InvalidGUID,
 
+    /// Malformed Page Allignment
```
tylerfanelli, line 83: s/Allignment/Alignment

src/firmware/host/types.rs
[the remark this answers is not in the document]
tylerfanelli: Why include it at all?

A suggestion: For the user API, get rid of this value altogether. When the API is issued, the library can traverse each cert in the `Vec<CertTableEntry>`, get the size of the cert, and add all of these sizes up (along with rounding up to the nearest page size). This sum will be the value in `certs_len` for the `SnpExtReport` request.

src/firmware/linux/guest/types.rs
```diff
@@ -68,33 +73,45 @@ pub struct SnpDerivedKeyRsp {
...
+            certs_address: vec![0u8; _4K_PAGE].as_mut_ptr() as u64,
+            certs_len: _4K_PAGE as u32,
         }
     }
+
+    /// This is used to update the certs_address offset when the buffer
+    /// needs to be extended.This happens when more than one page is required
+    /// for the storage.
+    pub fn extend_buffer(&mut self) {
+        self.certs_address = vec![0u8; self.certs_len as usize].as_mut_ptr() as u64;
```
tylerfanelli, line 107: Is this always guaranteed to be page aligned?

src/firmware/linux/guest/types.rs
```diff
@@ -347,6 +451,29 @@ bitfield! {
     pub single_socket_required, _: 20, 20;
 }
 
+impl Display for SnpGuestPolicy {
```
tylerfanelli, line 454: From clippy:
`using self as Display in impl Display will cause infinite recursion`

src/firmware/linux/host/types.rs
```diff
@@ -413,36 +439,48 @@ pub struct SnpSetExtConfig {
...
+    ) -> Result<Self, UserApiError> {
+        // Make sure the buffer is is of sufficient size.
+        if data.certs_buf < bytes.len() as u32 {
+            return Err(UserApiError::ApiError(SnpCertError::BufferOverflow));
+        }
+
+        // Make sure the buffer length is 4K-page aligned.
+        if data.certs_buf > 0 && data.certs_buf as usize % _4K_PAGE != 0 {
+            return Err(UserApiError::ApiError(SnpCertError::PageMisallignment));
+        }
```
tylerfanelli, line 463: From my above suggestion about getting rid of `certs_buf` in the user API struct, this check wouldn't be necessary. We could instead calculate the size of the certificates here.

tylerfanelli (commented):
Thanks Larry, I left some comments.

src/firmware/host/types.rs
```diff
@@ -87,6 +94,12 @@ impl std::fmt::Display for SnpCertError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
             SnpCertError::InvalidGUID => write!(f, "Invalid GUID provided in certificate chain."),
+            SnpCertError::PageMisallignment => {
+                write!(f, "Certificate Buffer not alligned with 4K Pages.")
```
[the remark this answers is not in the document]
larrydewey: TY. Resolved in next commit.

src/firmware/host/types.rs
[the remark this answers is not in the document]
larrydewey: @tylerfanelli It isn't in the user api, it is parsed from the user api into the kernel specific structure. 

@dubek Switched back to `certs_len` in next commit.

src/firmware/linux/host/types.rs
```diff
@@ -285,61 +284,83 @@ pub struct CertTableEntry {
...
+        for entry in table.iter() {
+            let guid: Uuid = match Uuid::parse_str(&entry.guid()) {
+                Ok(uuid) => uuid,
+                Err(_) => return Err(SnpCertError::InvalidGUID),
+            };
+
+            // Append the guid to the byte array.
+            guid.as_bytes().iter().for_each(|byte: &u8| {
+                bytes.push(*byte);
+            });
```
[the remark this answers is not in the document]
larrydewey: TY. Applied in next commit.

src/firmware/linux/host/types.rs
larrydewey: TY. Covered in next commit.

src/firmware/linux/host/types.rs
```diff
@@ -356,54 +377,59 @@ impl CertTable {
...
+            // we are finished.
+            retval.push(UapiCertTableEntry::from_guid(
+                &guid.hyphenated().to_string(),
+                cert_bytes.clone(),
+            ));
 
-            // Move the pointer ahead to the next value.
-            guid_ptr = guid_ptr.offset(ENTRY_SIZE);
+            // Cleanup the certificate vector for the next iteration.
+            cert_bytes.clear();
```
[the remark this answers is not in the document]
larrydewey: Done in the next commit.

src/firmware/linux/host/types.rs
```diff
@@ -356,54 +377,59 @@ impl CertTable {
...
+            }
 
-            // Calculate values for each CertTableEntry.
-            cert_address = guid_ptr.offset(OFFSET_START) as *mut u8;
-            cert_len = *(guid_ptr.offset(LENGTH_START) as *const usize);
-            cert_end = cert_address.add(cert_len);
-            guid = String::from_raw_parts(guid_ptr as *mut u8, GUID_SIZE, GUID_SIZE);
+            // Calculate the beginning and ending pointers of the raw certificate data.
+            cert_addr = table_ptr.offset(entry.offset as isize) as *mut u8;
+            cert_end = cert_addr.add(entry.length as usize) as *mut u8;
```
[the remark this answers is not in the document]
larrydewey: The bytes and pages should be verified when they are inserted. If we run into a scenario where the pages have been corrupted, this would be an issue. I am open to discussion about how to address this.

src/firmware/guest/mod.rs
larrydewey: I went with this approach in the end. Thank you for reminding me about it.

src/firmware/guest/mod.rs
larrydewey: TY. Next commit.

src/firmware/host/types.rs
larrydewey: TY. Fixed next commit.

src/firmware/host/types.rs
larrydewey: TY. Next commit.

src/firmware/linux/guest/types.rs
larrydewey: Not applicable anymore.

src/firmware/linux/host/types.rs
larrydewey: The `certs_buf` is only part of the FFI wrapper, and is required to communicate with the ioctl. We have to transform the user api into this.

src/firmware/host/types.rs
```diff
@@ -492,32 +505,50 @@ pub struct SnpExtConfig {
...
+
+impl SnpExtConfig {
+    /// Used to update the PSP with the cerificates provided.
+    pub fn update_certs_only(certificates: Vec<CertTableEntry>) -> Result<Self, SnpCertError> {
+        let mut certs_buffer: usize = _4K_PAGE;
+        let certs_length: usize = certificates.iter().map(|entry| entry.data().len()).sum();
+
+        while certs_length > certs_buffer {
+            certs_buffer += _4K_PAGE;
+        }
```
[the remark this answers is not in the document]
larrydewey: Fixed in next commit.

src/firmware/linux/host/types.rs
larrydewey: I am going to create a separate PR to add all of the unit tests so we aren't delaying the next release.

src/firmware/linux/host/types.rs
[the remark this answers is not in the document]
larrydewey: I think this is a good thing to bring up as an issue and we can address it there.

larrydewey:
@tylerfanelli this should be ready to go.

tylerfanelli (approved):
Great work!

PR #59, opened 2023-02-07
rustfmt: Add configuration file .rustfmt.toml

Zildj1an:
Add .rustfmt.toml configuration file that sets checking options for rustfmt and cargo fmt. Start basic configuration that enforces edition from 2021 and Unix style, which is the case for prior code history.

Signed-off-by: Carlos Bilbao <carlos.bilbao@amd.com>


tylerfanelli (approved):
Thanks!

PR #60, opened 2023-02-17
firmware/guest: Make FFICertTableEntry::parse_table() public

tylerfanelli:
I don't care much for `FFICertTableEntry` being publicly available to users, as it should never be used. However, the `parse_table()` method associated with the struct is quite useful for deserializing bytes into a cert table.

Maybe we should separate this function from `FFICertTableEntry` and only make `parse_table()` public?

### DIFF
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -5,12 +5,13 @@
 /// Rust-friendly types returned by FFI wrapping APIs.
 pub mod types;
 
+pub use super::linux::host::types::CertTableEntry as FFICertTableEntry;
+
 use std::fs::{File, OpenOptions};
 
 use self::types::*;
 
 use super::host::types::{CertTableEntry, Error, Indeterminate, UserApiError};
-use super::linux::host::types::CertTableEntry as FFICertTableEntry;
 use crate::firmware::linux::guest::ioctl::*;
 
 /// A handle to the SEV, SEV-ES, or SEV-SNP platform.
--- a/src/firmware/linux/host/types.rs
+++ b/src/firmware/linux/host/types.rs
@@ -366,6 +366,11 @@ impl CertTableEntry {
     /// };
     /// ```
     ///
+    /// # Safety
+    ///
+    /// Since this function manually parses the bytes of the table and utilizes pointers, it
+    /// violates the Rust memory safety guarentee.
+    ///
     pub unsafe fn parse_table(
         mut data: *mut CertTableEntry,
     ) -> Result<Vec<UapiCertTableEntry>, uuid::Error> {


larrydewey:
> I don't care much for `FFICertTableEntry` being publicly available to users, as it should never be used. However, the `parse_table()` method associated with the struct is quite useful for deserializing bytes into a cert table.

I agree, I would much rather not have this be publicly listed. Because this is a static method, we could just rip it out of the `impl` here, and re-export the function specificially via the `.../host/types.rs` module. I would prefer this to the guest module, as it is defined as a linux host type in the FFI.

> Maybe we should separate this function from `FFICertTableEntry` and only make `parse_table()` public?

That would make sense to me, with the exception I mentioned above.

tylerfanelli:
> > I don't care much for `FFICertTableEntry` being publicly available to users, as it should never be used. However, the `parse_table()` method associated with the struct is quite useful for deserializing bytes into a cert table.
> 
> I agree, I would much rather not have this be publicly listed. Because this is a static method, we could just rip it out of the `impl` here, and re-export the function specificially via the `.../host/types.rs` module. I would prefer this to the guest module, as it is defined as a linux host type in the FFI.
> 
> > Maybe we should separate this function from `FFICertTableEntry` and only make `parse_table()` public?
> 
> That would make sense to me, with the exception I mentioned above.

As I mentioned on our previous message, since `parse_table()` itself takes an `FFICertTableEntry` as an arg, we would need to publicize `FFICertTableEntry` anyway.

I could make a further note that from a user of this library, `FFICertTableEntry` should really only be used for its `parse_table()` method, and even then to exercise caution.

larrydewey:
We talked about this via Element, but just to document the discussion, there were a few ways to go about implementing the change without publicly listing the type. We are opting to go with a façade function and looking to create an enum to abstract what the user may send to that function and then casting those data pieces to the Linux FFI type before calling the internal function.

*Example:*
```rust
/// A data structure representing the possible data types which may be passed to
/// the parse_data function.
pub enum RawData {
    /// A pointer to a byte buffer
    Pointer(*mut u8),

    /// A heap allocation of bytes.
    Vector(Vec<u8>),
}

// ...

///
/// This is a façade for calling down to the private function which uses the Linux FFI wrapped type.
///
/// # Safety
///
/// This is unsafe because it calls an unsafe function to parse the Vec<CertTableEntry> from the data
/// which was provided.
pub unsafe fn parse_table(data: RawData) -> Result<Vec<CertTableEntry>, uuid::Error> {
    match data {
        RawData::Pointer(pointer) => {
            FFICertTableEntry::parse_table(pointer as *mut FFICertTableEntry)
        }
        RawData::Vector(vector) => {
            let mut copied_data = vector;
            FFICertTableEntry::parse_table(copied_data.as_mut_ptr() as *mut FFICertTableEntry)
        }
    }
}
```

larrydewey (changes requested):
See previous comments

tylerfanelli:
@larrydewey Yes, I see now. Will change.

tylerfanelli:
Closing this as after second thoughts, I think this should remain a private function.

PR #62, opened 2023-02-20
guest: Remove '&mut' from some public APIs

jepio:
Between v1.0 and v1.1, the public API changed to match the underlying ioctl API, which could potentially modify request objects. This is not necessary for the following functions, because:

- snp_get_report: the kernel does not write into the request object (and there are no fields for which it would make sense)
- snp_get_ext_report: the request object is copied into the extended request object expected by the ioctl. The kernel returns the certificate length in the request object but we hide this from the caller.
- snp_get_derived_key: the passed request object is copied before passing to the ioctl.

### DIFF
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -37,13 +37,13 @@ impl Firmware {
     pub fn snp_get_report(
         &mut self,
         message_version: Option<u8>,
-        report_request: &mut SnpReportReq,
+        mut report_request: SnpReportReq,
     ) -> Result<AttestationReport, Indeterminate<Error>> {
         let mut report_response: SnpReportRsp = SnpReportRsp::default();
 
         SNP_GET_REPORT.ioctl(
             &mut self.0,
-            &mut SnpGuestRequest::new(message_version, report_request, &mut report_response),
+            &mut SnpGuestRequest::new(message_version, &mut report_request, &mut report_response),
         )?;
 
         Ok(report_response.report)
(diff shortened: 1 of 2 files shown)


larrydewey (approved):
This looks good to me! Thanks!

larrydewey:
@tylerfanelli any thoughts?

src/firmware/guest/mod.rs
```diff
@@ -37,13 +37,13 @@ impl Firmware {
...
         message_version: Option<u8>,
-        report_request: &mut SnpReportReq,
+        mut report_request: SnpReportReq,
     ) -> Result<AttestationReport, Indeterminate<Error>> {
         let mut report_response: SnpReportRsp = SnpReportRsp::default();
 
         SNP_GET_REPORT.ioctl(
             &mut self.0,
-            &mut SnpGuestRequest::new(message_version, report_request, &mut report_response),
+            &mut SnpGuestRequest::new(message_version, &mut report_request, &mut report_response),
```
tylerfanelli, line 46: "_- snp_get_report: the kernel does not write into the request object (and there are no fields for which it would make sense)_"

You still give a `mut` reference here, however. Am I missing something here?

tylerfanelli (commented):
Thanks @jepio, I've just left one comment.

src/firmware/guest/mod.rs
jepio: No you're note missing anything. The ioctl still requires something potentially mutable to be passed in, but the kernel does not write into the request object for `SNP_GET_REPORT` (the `SnpReportReq` object does not have any output fields). I'm suggesting that we don't expose this internal property to the user, so that they do not need to create a mut variable themselves.

Issue #63, opened 2023-02-20
Change in Kernel User-space Behavior 


larrydewey:
The kernel patch by pgonda@google.com in kernel hash 47894e0f changed the ioctl return for SNP_GET_EXT_REPORT to succeed instead of returning an error when encountering an invalid certificate length. This was done to keep the cryptography safe. Because of this, any kernel before this hash (5.19, etc) behaves significantly differently, and we need to patch the way the API handles the error response.

tylerfanelli:
@larrydewey Can you link the commit with breakage?

larrydewey:
Hey @tylerfanelli I wasn't sure which link you were looking for:

[LKML Discussion and PR](https://lore.kernel.org/all/20221116175558.2373112-1-pgonda@google.com/)
[Github Context](https://github.com/torvalds/linux/commit/033e84da29cdee96cc2649504f05da5aace4b2df)

PR #64, opened 2023-02-20
63: Patching for change in userspace behavior.

larrydewey:
The kernel patch by pgonda@google.com in kernel hash 47894e0f changed the ioctl return for SNP_GET_EXT_REPORT to succeed instead of returning an error when encountering an invalid certificate length. This was done to keep the cryptography safe. Because of this, any kernel before this hash (5.19, etc) behaves differently than before, and we need to patch the way the API handles the error response.

resolves: #63

**EDIT 2023-02-21**
Adding Links for Context:
- [LKML Discussion and PR](https://lore.kernel.org/all/20221116175558.2373112-1-pgonda@google.com/)
- [Github Context](https://github.com/torvalds/linux/commit/033e84da29cdee96cc2649504f05da5aace4b2df)

### DIFF
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -109,14 +109,19 @@ impl Firmware {
                 &mut ext_report_response,
             );
 
+        // KEEP for Kernels before 47894e0f (5.19), as userspace broke at that hash.
+        if let Err(ioctl_error) = SNP_GET_EXT_REPORT.ioctl(&mut self.0, &mut guest_request) {
+            if guest_request.fw_err != INVALID_CERT_BUFFER {
+                return Err(ioctl_error.into());
+            }
+        }
+
         // The kernel patch by pgonda@google.com in kernel hash 47894e0f
         // changed the ioctl return to succeed instead of returning an
         // error when encountering an invalid certificate length. This was
         // done to keep the cryptography safe, so we will now just check
-        // the guest_request.fw_err for a new value. Any other ioctl errors
-        // need to be propagated up.
-        SNP_GET_EXT_REPORT.ioctl(&mut self.0, &mut guest_request)?;
-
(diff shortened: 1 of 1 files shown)


tylerfanelli (commented):
@larrydewey Hm. Only keeping for kernels before 47894e0f? How do we go about this? Is somebody supposed to remove this for kernels after 47894e0f?

larrydewey:
> @larrydewey Hm. Only keeping for kernels before 47894e0f? How do we go about this? Is somebody supposed to remove this for kernels after 47894e0f?

It is an error check which is only needed for kernels running before that hash was up-streamed; which happened in kernel v6.1. So everything between `v5.19.[0-17] - v6.0.[0-19]` will need the error check, but `>= v6.1` removes the error code response. We talked briefly about it on a call, but I have run through the kernel versions to isolate that it wasn't merged until exactly 6.1

src/firmware/guest/mod.rs
```diff
@@ -109,14 +109,19 @@ impl Firmware {
...
         // done to keep the cryptography safe, so we will now just check
-        // the guest_request.fw_err for a new value. Any other ioctl errors
-        // need to be propagated up.
-        SNP_GET_EXT_REPORT.ioctl(&mut self.0, &mut guest_request)?;
-
+        // the guest_request.fw_err for a new value.
+        //
         // Check to see if the buffer needs to be resized. If it does, the
         // we need to resize the buffer to the correct size, and
         // re-request for the certificates.
```
tylerfanelli, line 127: Wouldn't the `if` block below this comment have to be removed as well? We need to check the `certs_len` regardless? Or is the case that even though the `ioctl` returns a success, the `fw_err` is still set to `INVALID_CERT_BUFFER`?

tylerfanelli (approved):

Issue #33, opened 2022-10-24
Separate subcrate "snp"


tylerfanelli:
I've heard of interest between Enarx/Profian for a separate subcrate within this crate for all SNP-related data/code. I would assume this would be mostly utilizing the [snp](https://github.com/enarx/snp) crate and importing that crate in this library. Do I have this right, and can I get some more details of Enarx/Profian's plan here? I'm willing to work with others, just looking for some more clarity.

@larrydewey @npmccallum

tylerfanelli:
What would the policy be on crate ownership? I'd like to have commit/maintainer privileges on the github repo if possible.

npmccallum:
That's not a problem.

jepio:
So I think from a code management point of view it would be easier to keep two crates inside the same repository, but in different subdirectories. I don't know if rust has a problem with that layout, but that feels like the right layout, which would allow people to import snp without the rest of sev.

jepio:
Or to put things differently: in my opinion less-but-bigger repositories are easier to handle than more-but-smaller repositories.

tylerfanelli:
@jepio Yea, that's the way we're trying to go with this. Allowing others to just use the `snp` crate and ignore the rest of the legacy SEV is a goal.

See the [initial PR for the snp repo](https://github.com/virtee/snp/pull/1)

tylerfanelli:
@jepio Closing this. With our latest PR #74, you can now just use the SEV-SNP portion of the library by specifying this in a `Cargo.toml`:

`sev = { version = $VERSION, default-features = false, features = [ "snp" ] }`

PR #72, opened 2023-04-11
Remove SNP modules

tylerfanelli:
[virtee/snp](https://github.com/virtee/snp) now contains our SNP modules.


larrydewey (changes requested):
We will need to bump the version on this, but it all makes sense. What is our roadmap for incorporating the [snp](https://github.com/virtee/snp) crate into sev?

tylerfanelli:
Closes #33

tylerfanelli:
@larrydewey I've updated the version (yet haven't published it on crates).

fitzthum:
Have you considered using different rust features for the different functionality? I'm a bit skeptical about this being a breaking change.

fitzthum:
@tylerfanelli 

It is reasonable to separate SEV and SNP functionality, but to me creating two separate repositories (assuming that's the plan) is not idiomatic. Rust has a number of ways to handle this kind of situation. For instance, you can use workspaces to have multiple libraries within one repository or you can use features to have one library with several different configurations. Both of these approaches are more flexible, allowing you to easily create new libs/features, share code between components without any synchronization problems, etc. In contrast splitting things into a new repo seems a bit blunt.

This also creates an inconvenience for people who are already using the project (which has already had a major release). With features you can easily control what is the default functionality so you could avoid breaking changes for people who are already using the crate.

Finally, I'm not convinced that maintaining two crates separately would be easier. It seems like that would lead to fewer eyes on each, less mutual development, and more administrative overhead.

tylerfanelli:
@fitzthum 

Duly noted. Marking this as a draft for now and will explore using cargo workspaces.

PR #74, opened 2023-04-27
Conditional inclusion of SEV and SEV-SNP modules

tylerfanelli:


larrydewey (dismissed):
LGTM!

larrydewey:
It does look like `firmware::linux::host::types::snp::test::cert_table_entry::test_parse_table_offset_too_large` is failing in the CI, though?

tylerfanelli:
@larrydewey Yea, i'm not sure why that is, but I found the same problem on my local machine once (however, after another run, it was successful). Will investigate.

tylerfanelli:
@larrydewey The following tests are failing for me:
```
test snp::set_ext_config_cfg_only ... FAILED
test snp::set_ext_config_std ... FAILED
test snp::set_ext_invalid_config_std ... FAILED
```

I'm suspicious if this is due to our previous `set_ext_config` problems, especially because the certs only test succeeds:
`test snp::set_ext_config_certs_only ... ok`

Can you try this out for yourself?

tylerfanelli:
Can you re-review?

tylerfanelli:
@larrydewey Can you re-review?

tylerfanelli:
@fitzthum

What are your thoughts?

The final result: After this update, users will have to explicitly state which generation they would like in the library:

For an example `Cargo.toml`:

If user is looking for SEV and SEV-ES only:
`sev = { version = _, features = [ "sev" ] }`

SEV-SNP only:
`sev = { version = _, features = [ "snp" ] }`

Both:
`sev = { version = _, features = [ "sev",  "snp" ] }`

And thus, we allow for only specific generations to be included.

fitzthum (approved):
LGTM

I didn't check every change here, but the structural stuff looks very reasonable.

PR #75, opened 2023-05-05
snp_reset: Fixing the snp_reset API

larrydewey:
According to the specification, in order to reset the `REPORTED_TCB`, it is required that `snp_set_ext_config` receive a configuration with a TcbVersion of 0. Previously I was just sending a configuration address of zero, which does nothing.


tylerfanelli (approved):
@larrydewey 

LGTM. By any chance, could this be used to fix our `SET_EXT_CONFIG` tests? For instance, we could reset before the test.

PR #86, opened 2023-06-29
Legacy sev validation

larrydewey:
Adding functionality to verify Legacy SEV attestation reports are signed
by the PEK provided. This is required for an update to sevctl.


src/lib.rs
```diff
@@ -176,3 +184,48 @@ pub enum Generation {
...
+#[cfg(feature = "openssl")]
+impl From<Generation> for CertSevCaChain {
+    fn from(generation: Generation) -> CertSevCaChain {
+        use codicon::Decoder;
+
+        let (ark, ask) = match generation {
+            Generation::Naples => (SevBuiltin::naples::ARK, SevBuiltin::naples::ASK),
+            Generation::Rome => (SevBuiltin::rome::ARK, SevBuiltin::rome::ASK),
+            Generation::Milan => (SnpBuiltin::milan::ARK, SnpBuiltin::milan::ASK),
+            Generation::Genoa => (SnpBuiltin::genoa::ARK, SnpBuiltin::genoa::ASK),
```
DGonzalezVillal, line 197:
```
            Generation::Milan => (SnpBuiltin::milan::ARK, SnpBuiltin::milan::ASK),
            Generation::Genoa => (SnpBuiltin::genoa::ARK, SnpBuiltin::genoa::ASK),
```

We need to add legacy to certificates if we want to include SEV support for later generations.

src/firmware/guest/types/mod.rs
DGonzalezVillal, line 1: Do we want to add the legacy attestation report on the guest or host side of things?

DGonzalezVillal (changes requested):
Just a couple of comments on the legacy attestation report workflow.

src/firmware/guest/types/mod.rs
tylerfanelli: @DGonzalezVillal This is for legacy SEV, which doesn't include a `/dev/sev-guest` device for guests. This would only be for hosts.

tylerfanelli: Agreed. `firmware/guest` is only available with SEV-SNP at the moment.

src/firmware/guest/types/mod.rs
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
```
tylerfanelli, line 1: Currently, the `firmware/guest` module is only for SEV-SNP. This should be changed to allow users to include a legacy SEV attestation report type, correct?

src/certs/sev/mod.rs
```diff
@@ -53,8 +54,9 @@ pub trait Signer<T> {
     fn sign(&self, target: &mut T) -> Result<Self::Output>;
 }
 
+/// OpenSSL related signature
 #[cfg(feature = "openssl")]
-struct Signature {
+pub struct Signature {
```
tylerfanelli, line 59:
```suggestion
pub(crate) struct Signature {
```

src/certs/sev/mod.rs
```diff
@@ -71,14 +73,26 @@ pub struct PrivateKey<U> {
     usage: U,
 }
 
+/// Represents a public key.
 #[cfg(feature = "openssl")]
-struct PublicKey<U> {
+pub struct PublicKey<U> {
```
tylerfanelli, line 78:
```suggestion
pub(crate) struct PublicKey<U> {
```

src/certs/sev/mod.rs
```diff
@@ -29,8 +29,9 @@ use std::{
 #[cfg(feature = "openssl")]
 use openssl::*;
 
+/// OpenSSL body
 #[cfg(feature = "openssl")]
-struct Body;
+pub struct Body;
```
tylerfanelli, line 34:
```suggestion
pub(crate) struct Body;
```

src/lib.rs
```diff
@@ -56,6 +56,14 @@ use util::{TypeLoad, TypeSave};
 #[cfg(all(feature = "openssl", feature = "sev"))]
 use certs::sev::sev;
 
+#[cfg(feature = "openssl")]
+use certs::sev::ca::{Certificate, Chain as CertSevCaChain};
```
tylerfanelli, line 60: Since CA chains aren't changed between legacy SEV and SEV-SNP, we should extract the `ca` modules of `certs::sev` and `certs::snp` and only conditionally include what's different between the generations.

tylerfanelli (changes requested):

src/lib.rs
tylerfanelli: Can you elaborate?

larrydewey: Fixed in the next commit.

src/firmware/guest/types/mod.rs
larrydewey: Correct, that is what this would be doing. That being said, I wasn't sure where we should put this, as it is a guest-level action which will be consumed in `sevctl`. We could potentially just put it in the `host` module, but it isn't really a host action. I couldn't remember. Are we splitting it by device driver or by use-case?

src/certs/sev/mod.rs
larrydewey: Fixed in latest commits.

src/certs/sev/mod.rs
larrydewey: Fixed in latest commits.

src/certs/sev/mod.rs
larrydewey: Fixed in latest commits.

src/lib.rs
larrydewey: That is a good idea, but a bit outside of the scope of these changes. Do you want me to introduce those changes here, or should we make an issue and a new PR for it?

src/firmware/guest/types/mod.rs
larrydewey: We are re-exporting it into the types mod anyway. This shouldn't be an issue.

src/lib.rs
larrydewey: @tylerfanelli and I discussed this, it is an issue with PEM vs legacy cert format.

tylerfanelli (approved):

Issue #83, opened 2023-06-07
Unable to build with rust 1.65 - error[E0658]: custom discriminant values are not allowed in enums with tuple or struct variants


berrange:
Attempting to build sevctl on Ubuntu 22.04 distro fails with:

```
$ cargo build
   Compiling env_logger v0.10.0
   Compiling x509-parser v0.14.0
   Compiling sev v1.2.0
error[E0658]: custom discriminant values are not allowed in enums with tuple or struct variants
   --> /home/azureuser/.cargo/registry/src/github.com-1ecc6299db9ec823/sev-1.2.0/src/error.rs:366:20
    |
271 |     IoError(io::Error),
    |     ------------------ tuple variant defined here
...
366 |     RbModeExited = 0x001F, // 0x001F
    |                    ^^^^^^ disallowed custom discriminant
...
369 |     RMPInitRequired = 0x0020, // 0x0020
    |                       ^^^^^^ disallowed custom discriminant
    |
    = note: see issue #60553 <https://github.com/rust-lang/rust/issues/60553> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `sev` due to previous error
warning: build failed, waiting for other jobs to finish...
```

This is the OS info:

```
$ rustc --version
rustc 1.65.0
$ cargo --version
cargo 1.65.0
$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

IIUC, the error indicates sev is relying on a fairly new rust language feature which feels undesirable, when even the relatively modern Ubuntu 22.04 LTS release can't cope with it.

tylerfanelli:
I'm surprised this is disallowed, setting discriminate values for enums?

I guess a workaround we could do is remove the custom discriminant from the enum:
```
pub enum IoError {
    ..snip..
    RbModeExited,
    ...snip..
}
```

[...]

PR #89, opened 2023-07-17
Declare minimal Rust version in Cargo.toml.

sarutak:
This PR proposed to declare minimal Rust version in `Cargo.toml`.
Explicitly declaring the minimal version can help developers.

The version is considered what `cargo msrv` says.

```
$ cargo msrv
...
  Finished The MSRV is: 1.66.1 
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ readme = "README.md"
 keywords = ["amd", "sev"]
 categories = ["os", "os::linux-apis", "parsing", "network-programming", "hardware-support"]
 exclude = [ ".gitignore", ".github/*" ]
+rust-verson = "1.66.1"
 
 [badges]
 # See https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section


larrydewey:
Thanks for the work here @sarutak! @tylerfanelli I know we have discussed this a little bit. Are we still in agreement about changing the enums using custom discriminant values to lower the expected rust version dependency?

tylerfanelli:
@larrydewey #83 shows that a stock Ubuntu image uses a Rust version < 1.66. I feel that we should probably try to align the project with the "standard" Rust version on most distributions.

sarutak:
Hmm, Ubuntu 22.04 already provides Rust 1.66.1 right?
https://packages.ubuntu.com/jammy/devel/rustc

tylerfanelli:
My mistake, the issue I mentioned was for Rust 1.70, which isn't the standard on Ubuntu 22.04 yet.

tylerfanelli:
CC @larrydewey

Shouldn't this commit cause some issues due to the fact that the `Enum` mentioned in #83 uses features only available in Rust versions > 1.65?

tylerfanelli:
> I know this PR is setting it to 1.66, which I think still is broken using the custom discriminant feature, but maybe I had the release version wrong? Regardless, I think we were both in agreement that whatever the release version we set should remove the use of that feature, and the downgrade to allow older releases work. This is still our plan, correct?

I thought so too, but the `clippy` checks are still working?

larrydewey:
@tylerfanelli So it looks like version 1.66 was when the issue was patched. Any Rust version <= 1.65 will break. @sarutak where you made the PR for this, do you have any complaints with our downgrading the Rust version to at least 1.65?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ readme = "README.md"
 keywords = ["amd", "sev"]
 categories = ["os", "os::linux-apis", "parsing", "network-programming", "hardware-support"]
 exclude = [ ".gitignore", ".github/*" ]
+rust-verson = "1.66.1"
```
larrydewey, line 14:
```suggestion
rust-verson = "1.65.0"
```

larrydewey (changes requested):
If we are agreed, we will want to set the value to this, and implement the following changes:

From:
https://github.com/virtee/sev/blob/700b82054b73c186d89a393380768a60298fc275/src/error.rs#L360-L372

To:
```rust
...
 /// The AEAD algorithm would have overflowed 
 AEADOFlow, // 0x001D 
 
 /// A placeholder between AEDOFlow and RBModeExited
 _ReservedDoNotUse, // 0x001E
 /// A Mailbox mode command was sent while the SEV FW was in Ring Buffer 
 /// mode. Ring Buffer mode has been exited; the Mailbox mode command 
 /// has been ignored. Retry is recommended. 
 RbModeExited, // 0x001F 
  
 /// The RMP must be reinitialized. 
 RMPInitRequired, // 0x0020 
...
```

AND

From:
https://github.com/virtee/sev/blob/700b82054b73c186d89a393380768a60298fc275/src/error.rs#L430-L435

To:
```rust
...
             Error::AEADOFlow => "The AEAD algorithm would have overflowed.",
             Error::_ReservedDoNotUse => "If you are seeing this, something his HORRIBLY wrong!",
             Error::RbModeExited => "A Mailbox mode command was sent while the SEV FW was in Ring Buffer \ 
                                     mode. Ring Buffer mode has been exited; the Mailbox mode command has \ 
                                     been ignored. Retry is recommended.", 
...
```

FINALLY

From:
https://github.com/virtee/sev/blob/700b82054b73c186d89a393380768a60298fc275/src/error.rs#L533-L534

To:
```rust
...
             0x1D => Error::AEADOFlow,
             0x1E => Error::_ReservedDoNotUse,
             0x1F => Error::RbModeExited,
...
```

tylerfanelli:
LGTM.

DGonzalezVillal:
Hi, from what I'm understanding we want to downgrade the minimum version to 1.65 because it'll encompass most distributions shipped versions? What distros ship with rust version under 1.66? I've checked a few known distros and I've seen most of them ship with versions > 1.65.

larrydewey:
So after talking it over with Diego I deleted my previous comment. His point is very valid. it looks like even Ubuntu 20.04 created a security backport to upgrade Rust to 1.66.1, so at this point I think it would make the most sense to move forward with 1.66.1

larrydewey (approved):
LGTM!

tylerfanelli:
Agreed.

Fedora shipping 1.70: https://src.fedoraproject.org/rpms/rust

tylerfanelli (approved):

PR #94, opened 2023-08-09
launch/sev: separate KVM ENC region registration from update_data

FuuuOverclocking:
By registering KvmEncRegion, we can inform KVM about all memory owned by the SEV virtual machine. On the other hand, update_data can be utilized to encrypt the data we load into this memory. It's important to note that idle regions may not require encryption as they can be initialized by the VM's firmware or OS.

In the current API, these two operations are bundled together, resulting in unnecessary time consumption. Consequently, I have introduced two separate methods.

### DIFF
--- a/src/launch/sev.rs
+++ b/src/launch/sev.rs
@@ -105,6 +105,25 @@ impl<U: AsRawFd, V: AsRawFd> Launcher<Started, U, V> {
         Ok(())
     }
 
+    /// Register the encrypted memory region to a virtual machine.
+    /// Corresponds to the `KVM_MEMORY_ENCRYPT_REG_REGION` ioctl.
+    pub fn register_kvm_enc_region(&mut self, data: &[u8]) -> Result<()> {
+        KvmEncRegion::new(data).register(&mut self.vm_fd)?;
+        Ok(())
+    }
+
+    /// Encrypt guest data with its VEK, while the KVM encrypted memory region is not registered.
+    pub fn update_data_without_registration(&mut self, data: &[u8]) -> Result<()> {
+        let launch_update_data = LaunchUpdateData::new(data);
+        let mut cmd = Command::from(&self.sev, &launch_update_data);
+
+        LAUNCH_UPDATE_DATA
+            .ioctl(&mut self.vm_fd, &mut cmd)
+            .map_err(|e| cmd.encapsulate(e))?;
+
+        Ok(())
+    }
+
     /// Encrypt the VMSA on SEV-ES.
     pub fn update_vmsa(&mut self) -> Result<()> {
         let launch_update_vmsa = LaunchUpdateVmsa::new();


tylerfanelli:
@FuuuOverclocking Can you please address the `cargo clippy` issue and add a `Signed-off-by` line to the commit message?

FuuuOverclocking:
Fixed.

tylerfanelli:
> By registering KvmEncRegion, we can inform KVM about all memory owned by the SEV virtual machine. On the other hand, update_data can be utilized to encrypt the data we load into this memory. It's important to note that idle regions may not require encryption as they can be initialized by the VM's firmware or OS.
> 
> In the current API, these two operations are bundled together, resulting in unnecessary time consumption. Consequently, I have introduced two separate methods.

Can you elaborate a bit? Maybe a specific example where someone would want to register an encrypted region, yet not actually encrypt it with `SEV_LAUNCH_UPDATE`?

FuuuOverclocking:
> Can you elaborate a bit? Maybe a specific example where someone would want to register an encrypted region, yet not actually encrypt it with `SEV_LAUNCH_UPDATE`?

Sure. For instance, when I intend to start a VM with 2GB of memory, I would first register the kvm_userspace_memory_region. Following that, I would register the KVM ENC region of 2GB. Then load the kernel and firmware and use LAUNCH_UPDATE_DATA to encrypt them, which typically only amount to a few tens of MB.

tylerfanelli:
> > Can you elaborate a bit? Maybe a specific example where someone would want to register an encrypted region, yet not actually encrypt it with `SEV_LAUNCH_UPDATE`?
> 
> Sure. For instance, when I intend to start a VM with 2GB of memory, I would first register the kvm_userspace_memory_region. Following that, I would register the KVM ENC region of 2GB. Then load the kernel and firmware and use LAUNCH_UPDATE_DATA to encrypt them, which typically only amount to a few tens of MB.

So, in this case you are only actually encrypting a small subset of a VM's address space?

FuuuOverclocking:
> > > Can you elaborate a bit? Maybe a specific example where someone would want to register an encrypted region, yet not actually encrypt it with `SEV_LAUNCH_UPDATE`?
> > 
> > Sure. For instance, when I intend to start a VM with 2GB of memory, I would first register the kvm_userspace_memory_region. Following that, I would register the KVM ENC region of 2GB. Then load the kernel and firmware and use LAUNCH_UPDATE_DATA to encrypt them, which typically only amount to a few tens of MB.
> 
> So, in this case you are only actually encrypting a small subset of a VM's address space?

Yes, and I believe it is the usual usage, because encryption is quite time-consuming.

tylerfanelli:
CC @larrydewey What do you think of this?

larrydewey:
Where this is introducing a change around memory, I think it would be really beneficial to have some testing around this to make sure it doesn't break anything. Also, have you tested these changes on hardware?

larrydewey (approved):

tylerfanelli (approved):

PR #106, opened 2023-09-22
Modify workflow files to deal with warnings

sarutak:
GA workflow raises lots of warnings.
https://github.com/virtee/sev/actions/runs/6180387727

To deal with the warnings, I propose to update workflow files.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
     name: cargo clippy crypto_nossl
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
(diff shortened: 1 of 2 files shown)


larrydewey:
@sarutak Can you rebase this and we will review it?

sarutak:
@larrydewey Sure, I'll do it.

sarutak:
Checks have passed but it seems necessary to deal with some more warnings.
https://github.com/virtee/sev/actions/runs/7880457910
https://github.com/virtee/sev/actions/runs/7880457909
https://github.com/virtee/sev/actions/runs/7880457906

sarutak:
@larrydewey 
Now we have only one warning.
https://github.com/virtee/sev/actions/runs/7881703863

I think it's difficult to get rid of this warning for now because `actions-dco@1.1`(the current latest stable) and other replacements like [dco-check](https://github.com/christophebedard/dco-check) still use old Node.js.

`actions-rs/toolchain` is no longer maintained so it's better to replace it with `dtolnay/rust-toolchain` but how about leave `actions-dco` as is for now?

sarutak:
@larrydewey What do you think about the latest change?

PR #108, opened 2023-10-04
Build the C API with meson rather than cargo-c

tylerfanelli:
Rather than using `cargo-c` and building with `pkg-config`, we can directly incorporate the meson build system to build the library's C API. In doing this, projects using `meson` will be able to "wrap" the library and directly link with their projects.


build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+cd $(dirname $0)
+src=$(pwd)
+cd - >/dev/null
```
larrydewey, line 6: Adding in quotation to prevent splitting and `exit` in case the `cd` fails.
```suggestion
cd "$(dirname "$0")" || exit 1
src="$(pwd)"
cd - >/dev/null || exit 1
```

build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if [ ! -d build]; then
+    # configure a debug build (unoptimized and with debug info) for development
+    meson setup build --buildtype=debug
```
larrydewey, line 10: It would probably be a good idea to use the canonical path here `/usr/bin/meson` to prevent collision (not that it should happen, but still best practice). Also, we should probably include a check that `Meson` is installed, as well.

larrydewey (changes requested):

tylerfanelli:
@larrydewey Can you re-review?

larrydewey (approved):

Cargo.toml
```diff
@@ -21,13 +21,18 @@ maintenance = { status = "actively-developed" }
 is-it-maintained-issue-resolution = { repository = "virtee/sev" }
 is-it-maintained-open-issues = { repository = "virtee/sev" }
 
+[lib]
+crate-type = ["staticlib", "dylib", "rlib"]
```
cschramm, line 25: What's the idea of `dylib` here? Unfortunately it breaks builds with `panic="abort"`.

tylerfanelli: Good point, `dylib` is really for macOS which isn't really a concern here. I think it's safe to remove this. @larrydewey do you agree?

Issue #109, opened 2023-10-13
AttestationReport not verifiable without openssl feature


fspreiss:
The `sev::firmware::guest::types::snp::AttestationReport` is currently not verifiable without the `openssl` feature.

On the one hand, this is because `impl sev::certs::snp::Verifiable for (&Chain, &AttestationReport)` is only available with the `openssl` feature. On the other hand, even if one would want to implement the verification logic manually (for example with the pure-Rust `p384` crate), this is not possible because in the report's `signature: sev::certs::snp::ecdsa::Signature`, both `r` and `s` are private fields.

Are there plans to provide the verification logic also in a pure Rust fashion, that is, without the `openssl` feature?

If not, a stop-gap solution could be to create public getters for `r` and `s`, so that at least a manual verification (e.g., with [p384](https://crates.io/crates/p384)) becomes possible. Alternatively, one could make the fields `pub`lic, given that the attestation report's `signature` field is also public. Yet another alternative would be to provide a pure-rust `to_der` method, which is effectively what the `openssl`-gated `impl TryFrom<&Signature> for Vec<u8>` currently does.

In our project, we'd like to move away from OpenSSL so that all our code is in (safe) Rust. One of the last blockers in this regard is the verification of the attestation report, so it would be great if we could find a solution here.

In general, would you be open to contributions/PRs, e.g., for implementing the mentioned public getters for `r` and `s`?

larrydewey:
Closing, as your PR was merged :slightly_smiling_face:

fspreiss:
> Closing, as your PR was merged

@larrydewey, sure! Very happy to see the PR(s) merged!

Do you maybe have a rough estimate on when a new release would be cut that contains #111 (and #112)? This info would help us with our internal planning. Thanks.

PR #111, opened 2023-10-20
Allow verification of SNP AttestationReport without OpenSSL

fspreiss:
Allows the verification of an SNP `AttestationReport` without the use of OpenSSL through a new feature called `crypto_nossl`.

In particular, with `crypto_nossl` enabled, the `certs::snp::Verifiable` trait becomes implemented in pure Rust using the `p384`, `rsa`, `x509-cert`, and `sha2` crates, that is, without the use of OpenSSL, for the following types:
* `&certs::snp::Chain`
* `&certs::snp::ca::Chain`
* `(&certs::snp::Certificate, &certs::snp::Certificate)`
* `(&certs::snp::Chain, &firmware::guest::AttestationReport)`

This addresses https://github.com/virtee/sev/issues/109.

#### Tests and Github Workflows
* Adapts the feature gate for the SNP tests in `tests/certs.rs` so that those tests are run if either `openssl` or `crypto_nossl` is enabled.
* Adds negative tests for the SNP chain and report.
* Adds a `sw-crypto_nossl` test Github workflow to also perform the various cargo tests for the `crypto_nossl` feature.
* Splits the `cargo clippy` lint Github workflow into `cargo clippy openssl` and `cargo clippy crypto_nossl`, because currently the original workflow uses `--all-features`, which is not possible any more because `openssl` and `crypto_nossl` are mutually exclusive.

#### Feature & Compatibility
* `crypto_nossl` is not enabled by default and must be enabled manually if needed.
* `crypto_nossl` is _mutually exclusive_ with the existing `openssl` feature. If both are enabled at the same time, an explicit compiler error indicates that the two are incompatible. With this, a library user who wants to verify an SNP attestation report has to make a conscious decision as to whether OpenSSL shall be used for the verification or not. If OpenSSL shall be used, `openssl` is enabled. If it is OK not to use OpenSSL for verification, then `crypto_nossl` can be enabled.
* The feature name `crypto_nossl` is chosen deliberately as it makes it very explicit for library users that OpenSSL would _not_ be used for cryptographic verification.
* The existing OpenSSL-based implementation and API remains entirely unchanged. Existing library users are unaffected by this new feature, no matter if they currently have `openssl` enabled or not.

#### Misc
Removes the `pub use crate::firmware::host::Firmware;` re-export in `src/firmware/host/types/sev.rs` because this triggers an `error: unused import: crate::firmware::host::Firmware` in `src/firmware/host/types/sev.rs:6:9` in the clippy jobs. This seems to be a new clippy lint introduced in `1.75.0-nightly`, which is currently used by the Github workflow.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,6 +19,16 @@
 //!
 //! Refer to the [`launch`] module for more information.
 //!
+//! ## Cryptographic Verification
+//!
+//! To enable the cryptographic verification of certificate chains and
+//! attestation reports, either the `openssl` or `crypto_nossl` feature
+//! has to be enabled manually. With `openssl`, OpenSSL is used for the
+//! verification. With `crypto_nossl`, OpenSSL is _not_ used for the
+//! verification and instead pure-Rust libraries (e.g., `p384`, `rsa`,
+//! etc.) are used. `openssl` and `crypto_nossl` are mutually exclusive,
+//! and enabling both at the same time leads to a compiler error.
+//!
 //! ## Remarks
 //!
 //! Note that the Linux kernel provides access to these APIs through a set
@@ -45,6 +55,11 @@
 #![allow(unknown_lints)]
 #![allow(clippy::identity_op)]
 #![allow(clippy::unreadable_literal)]
+
+#[cfg(all(feature = "openssl", feature = "crypto_nossl"))]
+compile_error!(
+    "feature \"openssl\" and feature \"crypto_nossl\" cannot be enabled at the same time"
+);
 
 /// SEV and SEV-SNP certificates interface.
 pub mod certs;
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,8 +17,8 @@ jobs:
           command: fmt
           args: --all -- --check
 
-  clippy:
-    name: cargo clippy
+  clippy-openssl:
+    name: cargo clippy openssl
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -31,7 +31,23 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-features --all-targets -- -D clippy::all -D unused_imports -Dwarnings
+          args: --features=openssl,hw_tests,dangerous_hw_tests --all-targets -- -D clippy::all -D unused_imports -Dwarnings
+
+  clippy-crypto_nossl:
+    name: cargo clippy crypto_nossl
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          components: clippy
+          toolchain: nightly
+          profile: minimal
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --features=crypto_nossl,hw_tests,dangerous_hw_tests --all-targets -- -D clippy::all -D unused_imports -Dwarnings
 
   readme:
     name: cargo readme
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,8 @@
 on: [push, pull_request]
 name: test
 jobs:
-  sw:
-    name: sw ${{ matrix.toolchain }} ${{ matrix.profile.name }} ${{ matrix.features }}
+  sw-openssl:
+    name: sw openssl ${{ matrix.toolchain }} ${{ matrix.profile.name }} ${{ matrix.features }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -27,3 +27,30 @@ jobs:
             flag: --release
         features:
           - openssl
+
+  sw-crypto_nossl:
+    name: sw crypto_nossl ${{ matrix.toolchain }} ${{ matrix.profile.name }} ${{ matrix.features }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: ${{ matrix.profile.flag }} --features=${{ matrix.features }}
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain:
+          - nightly
+          - beta
+          - stable
+        profile:
+          - name: debug
+          - name: release
+            flag: --release
+        features:
+          - crypto_nossl
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ hw_tests = []
 dangerous_hw_tests = ["hw_tests"]
 sev = []
 snp = []
+crypto_nossl = ["dep:p384", "dep:rsa", "dep:sha2", "dep:x509-cert"]
 
 [target.'cfg(target_os = "linux")'.dependencies]
 iocuddle = "0.1"
@@ -61,6 +62,10 @@ bincode = "^1.3"
 hex = "0.4.3"
 libc = "0.2.147"
 lazy_static = "1.4.0"
+p384 = { version = "0.13.0", optional = true }
+rsa = { version = "0.9.2", optional = true }
+sha2 = { version = "0.10.8", optional = true }
+x509-cert = { version = "0.2.4", optional = true }
 
 [dev-dependencies]
 kvm-ioctls = ">=0.12"
--- a/README.md
+++ b/README.md
@@ -24,6 +24,16 @@ Refer to the [`firmware`] module for more information.
 
 Refer to the [`launch`] module for more information.
 
+### Cryptographic Verification
+
+To enable the cryptographic verification of certificate chains and
+attestation reports, either the `openssl` or `crypto_nossl` feature
+has to be enabled manually. With `openssl`, OpenSSL is used for the
+verification. With `crypto_nossl`, OpenSSL is _not_ used for the
+verification and instead pure-Rust libraries (e.g., `p384`, `rsa`,
+etc.) are used. `openssl` and `crypto_nossl` are mutually exclusive,
+and enabling both at the same time leads to a compiler error.
+
 ### Remarks
 
 Note that the Linux kernel provides access to these APIs through a set
--- a/src/certs/snp/builtin/genoa/mod.rs
+++ b/src/certs/snp/builtin/genoa/mod.rs
@@ -10,12 +10,12 @@ pub const ASK: &[u8] = include_bytes!("ask.pem");
 
 /// Get the Genoa ARK Certificate.
 pub fn ark() -> Result<Certificate> {
-    Ok(Certificate::from(X509::from_pem(ARK)?))
+    Certificate::from_pem(ARK)
 }
 
 /// Get the Genoa ASK Certificate.
 pub fn ask() -> Result<Certificate> {
-    Ok(Certificate::from(X509::from_pem(ASK)?))
+    Certificate::from_pem(ASK)
 }
 
 mod tests {
--- a/src/certs/snp/builtin/milan/mod.rs
+++ b/src/certs/snp/builtin/milan/mod.rs
@@ -10,12 +10,12 @@ pub const ASK: &[u8] = include_bytes!("ask.pem");
 
 /// Get the Milan ARK Certificate.
 pub fn ark() -> Result<Certificate> {
-    Ok(Certificate::from(X509::from_pem(ARK)?))
+    Certificate::from_pem(ARK)
 }
 
 /// Get the Milan ASK Certificate.
 pub fn ask() -> Result<Certificate> {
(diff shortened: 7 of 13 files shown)


larrydewey:
This is an interesting proposal. One of my hesitations with using the pure rust crypto libraries is that they have not been independently audited. Where there are some enterprise entities requiring FIPS certification, openssl made sense. How would you feel about adding in a separate conditional feature flag to opt for the pure rust libraries? @tylerfanelli and @DGonzalezVillal  any feelings about this?

DGonzalezVillal:
I understand wanting to be able to use the library without openssl, but then again we picked openssl because it's FIPS certified which a lot of people want/need. I think a conditional feature flag could be a good option to provide both solutions.

fspreiss:
Thanks for your comments, @larrydewey and @DGonzalezVillal, and for pointing out that some users may prefer a FIPS-certified verification library.

Based on your inputs, I adapted my proposal so that existing code isn't functionally changed, but additional code for verifying an SNP `AttestationReport` is added and gated with `#[cfg(not(feature = "openssl"))]` (note the `not`).

With this, an SNP report can be verified no matter if the optional, non-default `openssl` feature is enabled or not: if the `openssl` feature is enabled, OpenSSL is used. Otherwise, the pure-Rust implementation kicks in. This means all existing usages of the crate that have the `openssl` feature already enabled will continue to function in exactly the same way as before. Those who used the crate without the `openssl` feature so far, anyway didn't have the possibility to verify SNP reports, so also for them nothing changes, except that now they actually could verify SNP reports if they wanted to. Also, with this mechanism, there is no need for an additional feature-flag (to control if the pure-Rust implementation shall be used or not).

What do you think about this approach?

larrydewey:
> Thanks for your comments, @larrydewey and @DGonzalezVillal, and for pointing out that some users may prefer a FIPS-certified verification library.

No problem! Thank you for starting the PR!

> Based on your inputs, I adapted my proposal so that existing code isn't functionally changed, but additional code for verifying an SNP `AttestationReport` is added and gated with `#[cfg(not(feature = "openssl"))]` (note the `not`).
>
> With this, an SNP report can be verified no matter if the optional, non-default `openssl` feature is enabled or not: if the `openssl` feature is enabled, OpenSSL is used. Otherwise, the pure-Rust implementation kicks in. This means all existing usages of the crate that have the `openssl` feature already enabled will continue to function in exactly the same way as before. Those who used the crate without the `openssl` feature so far, anyway didn't have the possibility to verify SNP reports, so also for them nothing changes, except that now they actually could verify SNP reports if they wanted to. Also, with this mechanism, there is no need for an additional feature-flag (to control if the pure-Rust implementation shall be used or not).
> 
> What do you think about this approach?

I definitely like the idea of introducing the pure Rust solution for the cryptographic functionality, but I am a little hesitant to make it the default behavior of the application. My hesitation comes because I would be worried about someone setting the `openssl` feature flag, but then using the application thinking they are utilizing the library because the functionality is available. How would you feel about gating the functionality behind a new feature flag, and `or`-ing the logic so that only one of the options is viable at a time?

fspreiss:
> How would you feel about gating the functionality behind a new feature flag, and or-ing the logic so that only one of the options is viable at a time?

If I understand you correctly, you would like the new feature and `openssl` to be _mutually exclusive_. This is definitely doable. Note, however, that this is somewhat discouraged by the Cargo Book: see [this](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

To see how this would look like, the PR is now adapted so that the newly added code is gated with a new feature. It is called `pure_rust` for now: I'm not happy with the name, but the alternatives I could think of (e.g., `pure`, `verify`, `no_openssl`, etc.) also didn't seem much better. Maybe you have an idea. The PR description is not adapted yet. I'll adapt it once everyone is on board.

As recommended in the Rust book, I added the following:
```
#[cfg(all(feature = "openssl", feature = "purerust"))]
compile_error!("feature \"openssl\" and feature \"purerust\" cannot be enabled at the same time");
```
Note that in either case (that is, with or without the above compiler_error!), there are at least the following two compiler errors if both features are enabled at the same time: (1) `error[E0252]: the name Certificate is defined multiple times` and (2) `error[E0119]: conflicting implementations of trait certs::snp::Verifiable ...`. The additional, explicit compiler error just makes it clear what the problem is.

> My hesitation comes because I would be worried about someone setting the openssl feature flag, but then using the application thinking they are utilizing the library because the functionality is available.

What do you mean with "library" and with "is available" here? Do I understand you correctly that you are worried that someone setting the openssl feature flag could think that despite setting the openssl flag, the code would still not use OpenSSL for performing the verification, but the "pure rust" verification because this pure rust verification is also available without the openssl feature?

If yes, I'd say a simple and effective way to address this is through documentation. The README.md could say that by default the verification of an (SNP) chain and report is done purely in Rust, and that if someone wants this to be done by (the FIPS-certified) OpenSSL, the `openssl` feature has to be enabled. I'd say it is reasonable to assume that the documentation is consulted about the meaning of a feature before explicitly enabling this feature, especially if it is not enabled by default (which is the case here).
Another reason to make the pure-rust way the default is that the code becomes simpler and thus easier maintainable. For example, instead of being able to remove many `#[cfg(feature = "openssl")]`, with the new feature, those are now kept and turned into `#[cfg(any(feature = "openssl", feature = "pure_rust"))]`.

In any case, one argument that definitely speaks _for_ introducing a feature flag is that there is no good reason that someone who wants the verification done with OpenSSL has to pull in the various crates that are needed to do that (`p384`, `rsa`, `x509-cert`, etc.).

Let me know what you think.

fspreiss:
Update on the above comment: The `lint / cargo clippy` CI job is now failing because it enables `--all-features` which is of course not possible any more, so this would have to be adapted/fixed.

tylerfanelli:
Thanks @fspreiss, would you mind squashing the commits and adding a `Signed-off-by` line in the commit message?

tylerfanelli:
> > How would you feel about gating the functionality behind a new feature flag, and or-ing the logic so that only one of the options is viable at a time?
> 
> If I understand you correctly, you would like the new feature and `openssl` to be _mutually exclusive_. This is definitely doable. Note, however, that this is somewhat discouraged by the Cargo Book: see [this](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).
> 
> To see how this would look like, the PR is now adapted so that the newly added code is gated with a new feature. It is called `pure_rust` for now: I'm not happy with the name, but the alternatives I could think of (e.g., `pure`, `verify`, `no_openssl`, etc.) also didn't seem much better. Maybe you have an idea. The PR description is not adapted yet. I'll adapt it once everyone is on board.

I much prefer something like `crypto_nossl` or something similar.

> 
> As recommended in the Rust book, I added the following:
> 
> ```
> #[cfg(all(feature = "openssl", feature = "purerust"))]
> compile_error!("feature \"openssl\" and feature \"purerust\" cannot be enabled at the same time");
> ```
> 
> Note that in either case (that is, with or without the above compiler_error!), there are at least the following two compiler errors if both features are enabled at the same time: (1) `error[E0252]: the name Certificate is defined multiple times` and (2) `error[E0119]: conflicting implementations of trait certs::snp::Verifiable ...`. The additional, explicit compiler error just makes it clear what the problem is.
> 
> > My hesitation comes because I would be worried about someone setting the openssl feature flag, but then using the application thinking they are utilizing the library because the functionality is available.

I have this same concern. Right now, if a user would like to verify certificates without the `openssl` feature, it will simply fail and they will know to add the feature. With this, they will be verifying and potentially thinking they are using OpenSSL, but are not.

> 
> What do you mean with "library" and with "is available" here? Do I understand you correctly that you are worried that someone setting the openssl feature flag could think that despite setting the openssl flag, the code would still not use OpenSSL for performing the verification, but the "pure rust" verification because this pure rust verification is also available without the openssl feature?

I believe Larry is indicating my above point, where they will be unknowingly using the rust crypto solution instead of something like OpenSSL (which they will have _much_ more trust in).

> 
> If yes, I'd say a simple and effective way to address this is through documentation. The README.md could say that by default the verification of an (SNP) chain and report is done purely in Rust, and that if someone wants this to be done by (the FIPS-certified) OpenSSL, the `openssl` feature has to be enabled.

I disagree. I think that both should be optional features that need to be explicitly enabled.

> I'd say it is reasonable to assume that the documentation is consulted about the meaning of a feature before explicitly enabling this feature, especially if it is not enabled by default (which is the case here). Another reason to make the pure-rust way the default is that the code becomes simpler and thus easier maintainable. For example, instead of being able to remove many `#[cfg(feature = "openssl")]`, with the new feature, those are now kept and turned into `#[cfg(any(feature = "openssl", feature = "pure_rust"))]`.
> 
> In any case, one argument that definitely speaks _for_ introducing a feature flag is that there is no good reason that someone who wants the verification done with OpenSSL has to pull in the various crates that are needed to do that (`p384`, `rsa`, `x509-cert`, etc.).
> 
> Let me know what you think.

While I like this, I think that something like OpenSSL is watched and maintained very carefully, and that security-conscious users will definitely take issue with mistakenly using non-OpenSSL for crypto-related functionalty. Thus, I think that this should be a non-default feature.

fspreiss:
@tylerfanelli, thanks for your thoughts! What you write makes sense, and the new feature flag definitely seems the way to go here, so as to avoid any possible confusion regarding whether OpenSSL is used or not. With this in mind, the name `no_openssl` seems to make sense as feature name, so I renamed accordingly. I updated the PR description, fixed the CI issues, and squashed the commits with a sign-off. Please take a look.

DGonzalezVillal:
> I think that both should be optional features that need to be explicitly enabled.

I think this is the best solution. Making them mutually exclusive seems to be adding too much of a complicated layer.  If we make it so that the user has to at least enable 1, then the chance of confusion should be reduced, unless they for some reason decide to enable both at the same time. But I feel the library should still compile in that case.

DGonzalezVillal (approved):
code lgtm

src/lib.rs
```diff
@@ -46,6 +46,9 @@
 #![allow(clippy::identity_op)]
 #![allow(clippy::unreadable_literal)]
 
+#[cfg(all(feature = "openssl", feature = "no_openssl"))]
+compile_error!("feature \"openssl\" and feature \"no_openssl\" cannot be enabled at the same time");
```
DGonzalezVillal, line 50: Would there be a scenario were we want the ability to have both enabled? @larrydewey Or do we best keep them separate?

DGonzalezVillal (commented):
I

larrydewey (changes requested):
This looks great! My only ask would be to change out the generalized IO Errors into meaningful library errors (using error.rs).

tylerfanelli:
Small change, but I would prefer `crypto_nossl` instead of `no_openssl`.. a bit clearer.

tylerfanelli:
Also, a note in the README explaining the crypto_nossl feature would be useful.

fspreiss:
> Small change, but I would prefer crypto_nossl instead of no_openssl

> Also, a note in the README explaining the crypto_nossl feature would be useful.

@tylerfanelli, Done (for both of the above).

> My only ask would be to change out the generalized IO Errors into meaningful library errors (using error.rs).

@larrydewey, I'm happy to make some errors more specific, but I'm not sure I understand which exact errors you would prefer at which locations. There are the following locations in the new code where errors are used:

1. In `impl Verifiable for (&Chain, &AttestationReport)` and `impl Verifiable for (&Certificate, &Certificate)`

   Using a non-io-Error is not possible there, because the error type is determined by the `Verifiable` trait. Changing the actual trait definition seems somewhat out of scope, no?

2. In `impl TryFrom<&Signature> for p384::ecdsa::Signature`

   I'm using `io::Error` because this exact error is also used for all the other (three) `Signature`-related `TryFrom` conversions in that (`ecdsa`) module. It seemed to make sense to stay consistent with the rest of the code. If you prefer a different/dedicated error for this specific `TryFrom`, did you mean to re-use an existing error from `errors.rs` (if yes, which one), or did you mean to introduce a new one?

3. In the various methods in `impl Certificate` (`from_pem`, `to_pem`, `from_der`, `to_der`):

   I used the exact same API (incl. the same error types) that is used for the `openssl`-based implementation of `Certificate`. This way, library users can seamlessly switch between `openssl` and `crypto_nossl` without having to adapt any user-side code. If we were to switch away from `io::Error`, this seamless switch wouldn't be possible any more. Still, the errors there could be changed if you prefer that. If yes, did you mean to re-use an existing error from `errors.rs` (if yes, which one), or did you mean to introduce new ones?

larrydewey:
You bring up some good points. Let's not focus on those adjustments in this PR. However, @tylerfanelli it may be worth looking at not using a `std::io::Result`, as there are potentially other Errors which may come up from this. We can discuss it on the mailing list.

larrydewey (dismissed):

tylerfanelli (approved):

tylerfanelli:
@fspreiss Can you address the `Cargo.toml` conflict?

fspreiss:
> Can you address the Cargo.toml conflict?

@tylerfanelli, Done.

I had to force-push again so that everything is in a single signed-off commit. Looks like because of this the approvals need to be given again.

DGonzalezVillal (approved):
lgtm

larrydewey (approved):

Issue #122, opened 2023-11-13
Panic when getting an extended report


heavenboy8:
Hi,

I'm using the `sev` crate on an SEV-SNP VM provided by GCP (private preview). I want to generate an extended report using `pub fn get_ext_report([...])`. The code panics with that error : 

```
[...] panicked at 'misaligned pointer dereference: address must be a multiple of 0x4 but is 0xffffffffffffffff', /home/user/sev/src/firmware/linux/host/types/snp.rs:183:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
Aborted
```

The line 183 is `entry = *data;` from `parse_table` function called in the `get_ext_report` function. The value of `ext_report_request.certs_address` is equal to `u64::MAX` as it has been initialized by the constructor `pub fn new(data: &ReportReq)`. 

My thought is that on this machine no certificate has been cached by GCP and therefore no certificate was returned by the `ioctl`. The value of `ext_report_request.certs_len` is equal to zero before entering into the `unsafe` parsing instruction. 

I would like therefore to suggest a fix. Before parsing the `CertTableEntry`, check if the `ext_report_request.certs_len` is equal to zero. If so, we have 2 options:
- Return an `Err(CertError::EmptyCertBuffer)`
- Return an empty `Vec<CertTableEntry>`

As a user of the API, I guess the second option is better. If no certificate is configured, just return an empty `vec`. If we raise an error, I must implement an error handler which will call the non extend report  function whereas the report has already been generated by this previous function call. 

What do you think ?

heavenboy8:
@tylerfanelli any idea in mind concerning the issue I have reported?

Thanks again for the great job on this project.

tylerfanelli:
@heavenboy8 Sorry, just seeing this. I think option 2 would be best. Like you said, since there are no certs, just return an empty `Vec<CertTableEntry>`, or we could return an `Option<Vec<CertTableEntry>>` and return `None` with empty cert buffer.

tylerfanelli:
@heavenboy8 thanks!

larrydewey:
I am not sure I agree with @tylerfanelli  here. If you are calling for and extended request, but no certificates are loaded, I would think returning an error would make more sense, as we can specify more clearly that the certificate chain was not provided by the host. If we want to retrieve a standard attestation report, we should use:

[...]

PR #129, opened 2023-12-01
Fix panic when querying an extended report without certs

heavenboy8:
Related to https://github.com/virtee/sev/issues/122 discussed with @tylerfanelli 

Is is possible to review and merge please? 

Thanks

### DIFF
--- a/src/firmware/guest/mod.rs
+++ b/src/firmware/guest/mod.rs
@@ -122,7 +122,7 @@ impl Firmware {
         message_version: Option<u8>,
         data: Option<[u8; 64]>,
         vmpl: Option<u32>,
-    ) -> Result<(AttestationReport, Vec<CertTableEntry>), UserApiError> {
+    ) -> Result<(AttestationReport, Option<Vec<CertTableEntry>>), UserApiError> {
         let report_request = ReportReq::new(data, vmpl)?;
 
         let mut report_response = ReportRsp::default();
@@ -181,6 +181,10 @@ impl Firmware {
             return Err(UserApiError::FirmwareError(Error::InvalidConfig));
         }
 
+        if ext_report_request.certs_len == 0 {
+            return Ok((report_response.report, None));
+        }
+
         let certificates: Vec<CertTableEntry>;
 
         unsafe {
@@ -191,7 +195,7 @@ impl Firmware {
         }
 
         // Return both the Attestation Report, as well as the Cert Table.
-        Ok((report_response.report, certificates))
+        Ok((report_response.report, Some(certificates)))
     }
 
     /// Fetches a derived key from the AMD Secure Processor. The `message_version` will default to `1` if `None` is specified.


heavenboy8:
@larrydewey  any changes required regarding on the comments from the original issue #122 ?

tylerfanelli:
@heavenboy8 this looks fine to me. @larrydewey you had some concerns on #122, does you approve of this solution?

DGonzalezVillal:
@tylerfanelli,  @larrydewey  and I believe that an Error was more appropriate that a None/empty return. We think that if you are using the extended attestation report you are expecting some certificates to be there and if there are none, then you should be somehow warned. If you just want the attestation report you could just use the get_report(). Unless you think this type of error handling should be handled on the tooling side and not in the back-end api. We are open to hear your thoughts,

tylerfanelli:
> @tylerfanelli, @larrydewey and I believe that an Error was more appropriate that a None/empty return. We think that if you are using the extended attestation report you are expecting some certificates to be there and if there are none, then you should be somehow warned. If you just want the attestation report you could just use the get_report(). Unless you think this type of error handling should be handled on the tooling side and not in the back-end api. We are open to hear your thoughts,

I strongly disagree with this. For each of our APIs, they only fail if there is a problem with the underlying ioctl. Why should we diverge from this now with one API?

tylerfanelli:
> @DGonzalezVillal the strategy in my client code is:
> 
> * Ask for an extended report. Get the report and the certs
> * If the certs are not provided (which is the case on GCP) , then query AMD endpoints to get them
> 
> If the extended report function returns an error, in that case, then, I need to ask for a simple report and then query the AMD endpoints. It's quite a pity to fetch the simple report since the previous call got it but just returned an error (without the report). It's a useless ioctl call then.

Exactly. The behavior itself is fine (and successful). Error handling should be left for the case in which there's an actual error.

tylerfanelli:
> @tylerfanelli, @larrydewey and I believe that an Error was more appropriate that a None/empty return. We think that if you are using the extended attestation report you are expecting some certificates to be there and if there are none, then you should be somehow warned.

Isn't the `None` case warning enough? Your client code should see that the certs are `None`, and behave accordingly. What do we gain from an `Error` that `None` doesn't give us?

DGonzalezVillal:
@tylerfanelli Our train of thought is that if you're calling an extended attestation report you are **expecting** to get back both the report and the certificates. If you're getting an empty cert buffer then the host machine is in the wrong state and the extended report should fail. The extended report ioctl might be successful, but that only means that your request was successful, not so much that the machines are in the right state. Then it also adds the question of why even have a regular and extended attestation report as 2 separate commands, if they're now both essentially doing the same thing?

I also understand your point of view that the API should focus on the lower-level communication and the errors we bubble are strictly firmware, then let the user handle what the library returns however they want. But you do see how it is a little confusing as a user? If I'm requesting an extended report expecting certificates and I get an empty buffer, we don't necessarily know if it's that the host is hasn't cached certs or if there is some other issue. 

If you think this is the best way to handle empty certs, we can go with it. I just want to voice where we are coming from. If we merge this PR we will also have to bump release, since it does break API since we are now returning an option.

tylerfanelli:
> @tylerfanelli Our train of thought is that if you're calling an extended attestation report you are **expecting** to get back both the report and the certificates. If you're getting an empty cert buffer then the host machine is in the wrong state and the extended report should fail.

I don't feel that this assumption is correct. Take @heavenboy8's example that they mentioned above. They query if the certs are found locally; if not, they fetch from KDS.

 > The extended report ioctl might be successful, but that only means that your request was successful, not so much that the machines are in the right state.

I'm not too sure on this assumption either. You could be **querying** if the certs are there if a VM guest without prior knowledge of the system.

> Then it also adds the question of why even have a regular and extended attestation report as 2 separate commands, if they're now both essentially doing the same thing?

I think that's a legitimate question, and if we would like to remove `get_report`, then we could do so. `get_ext_report` does have the same functionality.

> 
> I also understand your point of view that the API should focus on the lower-level communication and the errors we bubble are strictly firmware, then let the user handle what the library returns however they want. But you do see how it is a little confusing as a user? If I'm requesting an extended report expecting certificates, the command works successfully, but I get an empty cert buffer, we don't necessarily know if it's that the host has not cached the certs or if there is some other issue.
> 

Right, but what if there actually are no certs cached? Then if we fail, we do not know if it's just because there are no certs or another issue.

Returning `None` would tell a user "_The command was successful, and there are no certificates currently cached_".

> If you think this is the best way to handle empty certs, we can go with it. I just want to voice where we are coming from. If we merge this PR we will also have to bump release, since it does break API since we are now returning an option.

Understood. Maybe we should cut a release before this is merged. We could then continue our other features/fixes that will also break the API (gmem comes to mind).

DGonzalezVillal:
@tylerfanelli Cool, I think then the None should work. Then we should handle the empty certs from tooling side. I also think leaving the regular get_report() should be fine, that way people don't have to handle certificates if they don't want to. I agree we should probably cut a release before merging this although.

DGonzalezVillal (approved):

heavenboy8:
up! ;)

heavenboy8:
Do you plan to merge this PR soon? How can I help @tylerfanelli ? 
Thanks again

tylerfanelli:
@heavenboy8 Apologies for the delay.

@DGonzalezVillal @larrydewey Should we merge this and update/increment the major version?

DGonzalezVillal:
I'm about to submit a PR for the ID-BLOCK calculation. Do you want to wait for that to be merged in too. Or we can merge this now and release the new version this week.

tylerfanelli:
@heavenboy8 Merged. We are going to wait a bit for another release. Do you mind using the upstream version for the time being?

heavenboy8:
> @heavenboy8 Merged. We are going to wait a bit for another release. Do you mind using the upstream version for the time being?

Thanks. I will do that indeed

PR #133, opened 2023-12-11
SEV generation string handling

tylerfanelli:


src/firmware/linux/host/types/snp.rs
```diff
@@ -108,7 +108,7 @@ impl CertTableEntry {
     /// ```
     ///
     #[cfg(target_os = "linux")]
-    pub fn uapi_to_vec_bytes(table: &Vec<UAPI::CertTableEntry>) -> Result<Vec<u8>, CertError> {
```
tylerfanelli, line 111: @larrydewey 

`clippy` suggested this change. Of course we want to follow its suggestions mostly, but this would break API... drop it for now? It provides no functional change and isn't exactly necessary.

DGonzalezVillal: I think the change should be fine, but if it is breaking API, maybe make it part of a different PR?

DGonzalezVillal (changes requested):

DGonzalezVillal:
Everything else looks good to me

tylerfanelli:
@DGonzalezVillal I removed the commit as I would like this merged without needing a major release. We can make the API breaking change on the next major release.

larrydewey:
@tylerfanelli Where we will be doing a major release anyway because of the kernel changes, should we look at just adding this in? @DGonzalezVillal is working on a PR to remove the unnecessary methods/functions which are no longer needed.

tylerfanelli:
@larrydewey By "this change", you're talking about [this](https://github.com/virtee/sev/pull/133#discussion_r1421954247) correct?

larrydewey:
I was referring more to your PR.

larrydewey:
@tylerfanelli Can you rebase and add support for Bergamo and Siena?

tylerfanelli:
Sure, will do.

tylerfanelli:
@larrydewey Added Bergamo + Siena support.

src/lib.rs
```diff
@@ -308,6 +332,55 @@ impl TryFrom<&sev::Chain> for Generation {
...
+
+            _ => Err(()),
+        }
+    }
+}
+
+#[cfg(any(feature = "sev", feature = "snp"))]
+impl Generation {
+    /// Create a title-cased string identifying the SEV generation.
+    pub fn titlecase(&self) -> String {
```
larrydewey, line 367: We should add Bergamo and Siena Support for these, too.

src/lib.rs
```diff
@@ -282,6 +288,24 @@ impl From<Generation> for CertSevCaChain {
     }
 }
 
+#[cfg(all(not(feature = "sev"), feature = "snp", feature = "openssl"))]
+impl From<Generation> for CertSnpCaChain {
+    fn from(gen: Generation) -> CertSnpCaChain {
```
larrydewey, line 293: We should add Bergamo and Siena Support for these, too.

larrydewey (changes requested):

larrydewey (approved):

DGonzalezVillal (approved):
lgtm

PR #139, opened 2024-01-17
Adding functions and types to calculate AUTH and ID blocks

DGonzalezVillal:
Adding functions and types that will allow users to calculate authority and id blocks


src/measurement/idblock.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Fill in an existing ID-Block with the provided parameters
+pub fn populate_auth_block(
+    id_auth: &mut IdAuth,
+    id_ec_pub_key: SevEcdsaPubKey,
+    author_pub_key: SevEcdsaPubKey,
+    id_sig: SevEcdsaSig,
+    author_sig: SevEcdsaSig,
+) -> Result<IdAuth, IdBlockError> {
+    id_auth.id_pubkey = id_ec_pub_key;
```
larrydewey, line 60: It might make more sense to have a constructor instead of a separate function?

src/measurement/idblock.rs
```diff
@@ -0,0 +1,133 @@
...
+    family_id: Option<FamilyId>,
+    image_id: Option<ImageId>,
+    svn: Option<u32>,
+    policy: Option<u64>,
+    id_key_file: PathBuf,
+    auth_key_file: PathBuf,
+) -> Result<IdMeasurements, IdBlockError> {
+    let mut measurements = IdMeasurements::default();
+
+    measurements.id_block = IdBlock::new(ld, family_id, image_id, svn, policy)?;
```
larrydewey, line 120: Same thing here, would it just be easier to create an object via constructor?

src/measurement/idblock_types.rs
```diff
@@ -0,0 +1,310 @@
...
+
+pub(crate) const ECDSA_PUBKEY_RESERVED: usize = 0x403 - 0x94 + 1;
+pub(crate) const ECDSA_SIG_RESERVED: usize = 0x1ff - 0x90 + 1;
+
+/// The expected launch digest of the guest
+#[repr(C)]
+#[derive(Default, Serialize, Deserialize)]
+pub struct IdBlockLaunchDigest(LargeArray<u8, ID_BLK_DIGEST_BYTES>);
+
+impl TryFrom<Vec<u8>> for IdBlockLaunchDigest {
```
larrydewey, line 47: Would it make more sense to implement for `&[u8]` here, instead?

larrydewey (changes requested):
Some discussion points, and you have a number of linting errors which need to be addressed.

src/measurement/idblock.rs
```diff
@@ -0,0 +1,118 @@
...
+        id_sig,
+        id_ec_pub_key,
+        author_sig,
+        author_pub_key,
+    ))
+}
+
+///Read a pem key file and return a private EcKey.
+/// Key has to be an EC P-384 key.
+pub fn load_priv_key_from_pem(path: PathBuf) -> Result<EcKey<Private>, IdBlockError> {
```
larrydewey, line 58: Do we foresee any requests to want DER support?

src/measurement/vmsa.rs
```diff
@@ -113,6 +113,71 @@ where
     }
 }
 
+impl<T, const N: usize> TryFrom<Vec<T>> for LargeArray<T, N>
```
larrydewey, line 116: Do we want to implement this for Vec, or should we adjust this to do it to a slice?
```suggestion
impl<T, const N: usize> TryFrom<&[T]> for LargeArray<T, N>
```
We will still want to clone/copy the data, but it may save a move action.

src/measurement/vmsa.rs
```diff
@@ -113,6 +113,71 @@ where
...
+where
+    T: std::marker::Copy + std::default::Default + for<'a> Deserialize<'a> + Serialize,
+{
+    type Error = LargeArrayError;
+
+    fn try_from(vec: Vec<T>) -> Result<Self, Self::Error> {
```
larrydewey, line 122: If we go with the top thing...

```suggestion
    fn try_from(vec: &[T]) -> Result<Self, Self::Error> {
```

src/measurement/vmsa.rs
```diff
@@ -113,6 +113,71 @@ where
...
+    T: std::marker::Copy + std::default::Default + for<'a> Deserialize<'a> + Serialize,
+{
+    type Error = LargeArrayError;
+
+    fn try_from(array: [T; N]) -> Result<Self, Self::Error> {
+        Ok(LargeArray(array))
+    }
+}
+
+impl<T, const N: usize> TryFrom<&[u8]> for LargeArray<T, N>
```
larrydewey, line 146: If we go with the top instance, we wouldn't need this implementation.

src/measurement/vmsa.rs
```diff
@@ -113,6 +113,71 @@ where
...
+        }
+    }
+}
+
+impl<T, const N: usize> LargeArray<T, N>
+where
+    T: std::marker::Copy + std::default::Default + for<'a> Deserialize<'a> + Serialize,
+{
+    /// Get the large array as a regular array format
+    pub fn as_array(&self) -> [T; N] {
```
larrydewey, line 176: I like this, especially if they want a copy. It may be worth looking into returning a slice, as well?

larrydewey (commented):
Just a couple thoughts. Not blockers, but something to think about.

DGonzalezVillal:
Also added the Guest Features for the VMSA to this PR. This can affect the measurement depending on which kernel features are enabled.

src/measurement/idblock.rs
DGonzalezVillal: yes :(

Begin ec private key

larrydewey: ----- BEGIN PUBLIC KEY -----

You mean this? ^^

src/error.rs
```diff
@@ -738,6 +738,104 @@ impl std::fmt::Display for SevHashError {
...
+
+    /// Error when converting from vector
+    VectorError(String),
+}
+
+impl std::fmt::Display for LargeArrayError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            LargeArrayError::SliceError(error) => {
+                write!(f, "Error when trying form slice: {error}")
```
larrydewey, line 755:
```suggestion
                write!(f, "Error when trying from slice: {error}")
```

src/error.rs
```diff
@@ -738,6 +738,104 @@ impl std::fmt::Display for SevHashError {
...
+            }
+            LargeArrayError::VectorError(error) => {
+                write!(f, "Error when trying form vector: {error}")
```
larrydewey, line 758:
```suggestion
                write!(f, "Error when trying from vector: {error}")
```

larrydewey (changes requested):

DGonzalezVillal:
@tylerfanelli whenever you can, can you please review

larrydewey (approved):
LGTM

src/error.rs
```diff
@@ -740,6 +740,104 @@ impl std::fmt::Display for SevHashError {
...
+impl std::convert::From<std::io::Error> for IdBlockError {
+    fn from(value: std::io::Error) -> Self {
+        Self::FileError(value)
+    }
+}
+impl std::convert::From<bincode::ErrorKind> for IdBlockError {
+    fn from(value: bincode::ErrorKind) -> Self {
+        Self::BincodeError(value)
+    }
+}
```
tylerfanelli, line 839: Can't these be converted by `map_err()`? Is the `From` implementation necessary?

tylerfanelli (approved):
One small comment. Other than that, LGTM.

PR #142, opened 2024-01-24
Changing API for kernel 6.6 new IOCTL structure

DGonzalezVillal:
In new kernel 6.6 the SNP IOCTL structure changed. We are completely removing the deprecated extended config and IOCTLs. I also bumped the release to 3.0 in this PR. THIS WILL NOT WORK ON KERNELS BEFORE SNP 6.6.

The new IOCTLS and it's uses have been added to this PR. 

SNP_SET_CONFIG_START and SNP_SET_CONFIG_END will be revised in later kernel patches according to Michael Roth. So those will probably be changed. Should we include those in this PR? 

Also certificates are no longer handled by firmware, but they will be handled by the QEMU command line. Do we want to move the kernel version of CertTableEntry to the UAPI? that way users can create the appropriate versions of that? For now there will be some compiling errors because that struct is not being used, but I don't think it has been deprecated yet. 

Everything related to the previous ioctls and extended_config has been deleted.

Please share thoughts and comments on the topic.

### DIFF
--- a/src/firmware/host/mod.rs
+++ b/src/firmware/host/mod.rs
@@ -196,7 +196,11 @@ impl Firmware {
         Ok(platform_status)
     }
 
-    /// Commit the current SNP firmware
+    /// The firmware will perform the following actions:  
+    /// - Set the CommittedTCB to the CurrentTCB of the current firmware.  
+    /// - Set the CommittedVersion to the FirmwareVersion of the current firmware.  
+    /// - Sets the ReportedTCB to the CurrentTCB.  
+    /// - Deletes the VLEK hashstick if the ReportedTCB changed.
     ///
     /// # Example:
     /// ```ignore
@@ -222,57 +226,16 @@ impl Firmware {
     /// );
     /// let mut firmware: Firmware = Firmware::open().unwrap();
     ///
-    /// let status: bool = firmware.set_ext_config(configuration).unwrap();
+    /// let status: bool = firmware.snp_set_config(configuration).unwrap();
     /// ```
     #[cfg(feature = "snp")]
     pub fn snp_set_config(&mut self, new_config: Config) -> Result<(), UserApiError> {
-
-        SNP_SET_CONFIG.ioctl(&mut self.0, &mut Command::from_mut(&mut new_config.try_into()?))?;
+        SNP_SET_CONFIG.ioctl(
+            &mut self.0,
+            &mut Command::from_mut(&mut new_config.try_into()?),
+        )?;
 
         Ok(())
-    }
-
-    /// Start SNP configuration process
-    ///
-    /// # Example:
-    /// ```ignore
-    /// let mut firmware: Firmware = Firmware::open().unwrap();
-    ///
-    /// let config_id: configTransaction = firmware.snp_set_config_start().unwrap();
-    /// ```
-    #[cfg(feature = "snp")]
-    pub fn snp_set_config_start(&mut self) -> Result<ConfigTransaction, UserApiError> {
-
-        let mut config_start: FFI::types::SnpSetConfigStart = Default::default();
-
-        SNP_SET_CONFIG_START.ioctl(&mut self.0, &mut Command::from_mut(&mut config_start))?;
-
-        Ok(config_start.try_into()?)
-    }
-
-    /// End SNP configuration process
-    ///
-    /// # Example:
-    /// ```ignore
-    /// let mut firmware: Firmware = Firmware::open().unwrap();
-    ///
-    /// let start_config_id: configTransaction = firmware.snp_set_config_start().unwrap();
-    /// 
-    /// let end_config_id: configTransaction = firmware.snp_set_config_end().unwrap();
-    /// 
-    /// // if start and end error id's don't match, assume something process failed and start again. 
-    /// if start_config_id != end_config_id {
-    ///     Err(eprintln!("start id and end id don't match!""))
-    /// }
-    /// ```
-    #[cfg(feature = "snp")]
-    pub fn snp_set_config_end(&mut self) -> Result<ConfigTransaction, UserApiError> {
-        
-        let mut config_end: FFI::types::SnpSetConfigEnd = Default::default();
-
-        SNP_SET_CONFIG_END.ioctl(&mut self.0, &mut Command::from_mut(&mut config_end))?;
-
-        Ok(config_end.try_into()?)
     }
 }
 
--- a/src/firmware/host/mod.rs
+++ b/src/firmware/host/mod.rs
@@ -38,6 +38,10 @@ use std::mem::MaybeUninit;
 #[cfg(target_os = "linux")]
 use std::convert::TryInto;
 
+#[cfg(feature = "snp")]
+#[cfg(target_os = "linux")]
+use super::linux::host::types::SnpCommit;
+
 /// The CPU-unique identifier for the platform.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Identifier(pub Vec<u8>);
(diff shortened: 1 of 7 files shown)


src/firmware/linux/host/types/snp.rs
```diff
@@ -211,34 +211,61 @@ impl CertTableEntry {
...
-    /// previous certificate(s) should be removed via SNP_SET_EXT_CONFIG.
-    pub certs_address: u64,
+/// SNP_COMMIT structure
+#[cfg(feature = "snp")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
+#[repr(C,packed)]
+pub struct SnpCommit{
+    /// len of the command buffer read by the PSP
+    length: u32
+}
```
larrydewey, line 221: Again, may be worth looking at doing this?
```suggestion
/// SNP_COMMIT structure
/// - length: length of the command buffer read by the PSP
#[cfg(feature = "snp")]
#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
#[repr(C,packed)]
pub struct SnpCommit(u32);
```

src/firmware/linux/host/types/snp.rs
```diff
@@ -211,34 +211,61 @@ impl CertTableEntry {
...
+/// Sets the system wide configuration values for SNP.
+#[cfg(feature = "snp")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(C, packed)]
+pub struct SnpSetConfig {
+    /// The TCB_VERSION to report in guest attestation reports.
+    pub reported_tcb: UAPI::TcbVersion,
+
+    /// mask_chip_id [0] : whether chip id is present in attestation reports or not
+    /// mask_chip_key [1]: whether attestation reports are signed or not
```
larrydewey, line 232:
```suggestion
    /// mask_id [0] : whether chip id is present in attestation reports or not
    /// mask_id [1]: whether attestation reports are signed or not
```

src/firmware/linux/host/types/snp.rs
```diff
@@ -211,34 +211,61 @@ impl CertTableEntry {
...
-    /// 4K-page aligned length of the buffer which will hold the fetched certificates.
-    pub certs_len: u32,
+/// Metadata for config transactions.
+#[cfg(feature = "snp")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
+#[repr(C, packed)]
+pub struct SnpSetConfigEnd {
+    /// The ID of the transaction ended by a call SNP_SET_CONFIG_END.
+    pub id: u64
 }
```
larrydewey, line 266: Like above
```suggestion
/// Metadata for config transactions.
/// Contains the ID of the transaction ended by a call SNP_SET_CONFIG_END.
#[cfg(feature = "snp")]
#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
#[repr(C, packed)]
pub struct SnpSetConfigEnd(u64);
```

larrydewey (changes requested):

src/firmware/linux/host/ioctl.rs
```diff
@@ -100,16 +106,29 @@ pub const GET_ID: Ioctl<WriteRead, &Command<GetId<'_>>> = unsafe { SEV.write_rea
...
+/// Commit the currently installed firmware.
+/// C IOCTL calls -> sev_ioctl_do_snp_commit
 #[cfg(feature = "snp")]
-pub const SNP_SET_EXT_CONFIG: Ioctl<WriteRead, &Command<SnpSetExtConfig>> =
-    unsafe { SEV.write_read(0) };
+pub const SNP_COMMIT: Ioctl<WriteRead, &Command<SnpCommit>> =
+    unsafe {SEV.write_read(0)};
 
-/// Get the SNP Extended Configuration Settings.
+/// Set the system-wide configuration such as reported TCB version in the attestation report
```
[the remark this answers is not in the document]
DGonzalezVillal: Waiting until SNP_CONFIG_START and END are updated in kernel

src/firmware/linux/host/types/snp.rs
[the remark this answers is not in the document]
DGonzalezVillal: Keeping it since users still need to provide GHCB spec certs to QEMU
https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56421.pdf

src/firmware/host/mod.rs
```diff
@@ -192,107 +196,83 @@ impl Firmware {
...
     /// # Example:
     /// ```ignore
+    /// let configuration = Config::new(
+    ///     TcbVersion::new(3, 0, 10, 169),
+    ///     0,
+    /// );
     /// let mut firmware: Firmware = Firmware::open().unwrap();
     ///
-    /// let status: ExtConfig = firmware.get_ext_config().unwrap();
+    /// let status: bool = firmware.set_ext_config(configuration).unwrap();
```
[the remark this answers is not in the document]
DGonzalezVillal: Waiting until snp_start and end are finished on the kernel side

src/firmware/host/types/snp.rs
```diff
@@ -429,3 +289,65 @@ impl TcbVersion {
...
+#[cfg(feature = "snp")]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ConfigTransaction {
+    /// START:
+    ///     The ID of the transaction started by a call to SNP_SET_CONFIG_START.
+    ///
+    /// END:
+    ///     The ID of the transaction ended by a call SNP_SET_CONFIG_END.
+    pub id: u64
+}
```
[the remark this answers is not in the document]
DGonzalezVillal: Removing until start and end are in a more finalized state

DGonzalezVillal:
Removed SNP_CONFIG_START and SNP_CONFIG_END since we are expecting to have big changes to those in the future. We will can use SNP_COMMIT and SNP_SET_CONFIG as they are right now.

src/firmware/host/types/snp.rs
```diff
@@ -122,6 +124,11 @@ impl CertTableEntry {
     pub fn new(cert_type: CertType, data: Vec<u8>) -> Self {
         Self { cert_type, data }
     }
+
+    /// Builds a Kernel formatted CertTable for sending the certificate content to the PSP.
+    pub fn cert_table_to_vec_bytes(table: &Vec<Self>) -> Result<Vec<u8>, CertError> {
```
larrydewey, line 129:
```suggestion
    pub fn cert_table_to_vec_bytes(table: &[Self]) -> Result<Vec<u8>, CertError> {
```

larrydewey: Can we make this adjustment? I believe this was one of the linting issues.

larrydewey (changes requested):
I **_THINK_** this is the last comment :)

larrydewey (dismissed):

larrydewey (approved):
LGTM! @tylerfanelli any additional thoughts about this?

tylerfanelli:
@larrydewey I don't have an updated kernel to test this with at the moment. Is `snp-latest` the branch with these changes?

larrydewey:
@tylerfanelli Yes, this is for the snp-latest branch. Diego is testing it on a system right now. Almost done.

src/firmware/host/types/snp.rs
```diff
@@ -122,6 +126,19 @@ impl CertTableEntry {
     pub fn new(cert_type: CertType, data: Vec<u8>) -> Self {
         Self { cert_type, data }
     }
+
+    /// Builds a Kernel formatted CertTable for sending the certificate content to the PSP.
+    pub fn cert_table_to_vec_bytes(table: &[Self]) -> Result<Vec<u8>, CertError> {
+        FFI::types::CertTableEntry::uapi_to_vec_bytes(&table.to_vec())
```
larrydewey, line 132: Didn't we plan to update `uapi_to_vec_bytes` to receive a reference instead of a vector? If so, we will need to adjust this to not be sending a vector.
```suggestion
        FFI::types::CertTableEntry::uapi_to_vec_bytes(table)
```

src/firmware/host/types/snp.rs
```diff
@@ -122,6 +126,19 @@ impl CertTableEntry {
...
+    }
+
+    /// Takes in bytes in kernel CertTable format and returns in user API CertTable format.
+    pub fn vec_bytes_to_cert_table(mut bytes: Vec<u8>) -> Result<Vec<Self>, CertError> {
```
larrydewey, line 136: We should probably just pass this by reference, so we aren't doing a move/copy here.

```suggestion
    pub fn vec_bytes_to_cert_table(bytes: &mut [u8]) -> Result<Vec<Self>, CertError> {
```

larrydewey (changes requested):

larrydewey (approved):

PR #151, opened 2024-02-12
Certs: Adding generic support.

larrydewey:
Adding generic support to automatically identify which type of certificate the user provided and construct a Certificate from the identity discovered.

Also adds an explicit `openssl` feature flag instead of relying on the inferred flag.

### DIFF
--- a/src/certs/snp/cert.rs
+++ b/src/certs/snp/cert.rs
@@ -2,6 +2,7 @@
 
 use super::*;
 
+use crate::error::CertFormatError;
 use openssl::pkey::{PKey, Public};
 use openssl::x509::X509;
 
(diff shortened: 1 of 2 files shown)


src/certs/snp/cert.rs
```diff
@@ -10,6 +10,49 @@ use openssl::x509::X509;
...
+
+impl std::error::Error for CertFormatError {}
+
+impl std::fmt::Display for CertFormatError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::UnknownFormat => write!(f, "Unknown Certificate Format Encountered."),
+        }
+    }
+}
```
DGonzalezVillal, line 26: Do we want this in the error file instead?

DGonzalezVillal (changes requested):
Logic looks good, just an organization tidbit

Cargo.toml
```diff
@@ -42,6 +42,7 @@ hw_tests = []
 dangerous_hw_tests = ["hw_tests"]
 sev = []
 snp = []
+openssl = ["dep:openssl"]
```
DGonzalezVillal, line 45: Would we like to mention in the commit message that openssl is being added as a feature in this PR?

DGonzalezVillal (approved):
Just a question regarding adding OpenSSL as a feature, but besides that the code looks good to me.

Cargo.toml
tylerfanelli: Wasn't it already a feature? What does this line do?

larrydewey: @tylerfanelli this makes it an explicit feature defined for the library.

tylerfanelli (approved):

PR #153, opened 2024-02-16
152: Fixing cert identifier bug

larrydewey:
There was a bug where we were comparing the whole
byte stream instead of the first 27 bytes of data.


DGonzalezVillal (approved):
fixed!

PR #154, opened 2024-02-16
snp-certs: Adding unit tests

larrydewey:
Adding some unit tests for the `identify_format(..)` function to make sure that it doesn't get broken through modification in the future.


DGonzalezVillal (approved):
lgtm

Issue #165, opened 2024-03-21
Inconsistency in guest policy structs used for attestation and id block


its-luca:
There seem to be at least two structs [2] and [3] that model the guest policy defined in Table 9 in [1]. However the definition in [3] only defines a subset of the features defined in [2]. 

Besides this discrepancy, the fact that there are two structs for the guest policy is quite confusing, as [3] is used when generating the ID block and [3] is used for the attestation report. Thus when writing software to start a SEV VM and do attestation we need to input the same information in two different formats/structs. It would seem more natural to me to use one definition.

If you agree with this assessment, I could look into refactoring the code accordingly.

[1] https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf
[2] https://github.com/virtee/sev/blob/0e15b313691ea32029e579fa59a4e9d399fd0a1d/src/firmware/guest/types/snp.rs#L420
[3] https://github.com/virtee/sev/blob/0e15b313691ea32029e579fa59a4e9d399fd0a1d/src/launch/snp.rs#L129

larrydewey:
@its-luca This is definitely a duplication of code. In my opinion, it would make the most sense to import the definition from `sev/src/firmware/guest/types/snp.rs` into `sev/src/launch/snp.rs`. @tylerfanelli do you have any opinion about this?

tylerfanelli:
Importing the definition makes the most sense, I would agree.

PR #168, opened 2024-03-30
Unify Duplicated GuestPolicy Definitions

its-luca:
Unifies the duplicated `GuestPolicy` definition, as discussed in https://github.com/virtee/sev/issues/165  .

I am not quite happy with the "ergonomics" of the bitflags! macro. If I am not overlooking something, it requires to instantiate a mutable variable and use the set_XXX functions to create an instance using the bit definitions. Crates like [modular_bitfield](https://docs.rs/modular-bitfield/0.11.2/modular_bitfield/) seem more ergonomic, but I don't have a good overview of the available crates for this problem space


DGonzalezVillal (approved):
The code LGTM. This will be an API change, so we would need to bump version and release after this merge.

larrydewey (approved):

PR #169, opened 2024-03-30
Fix broken documentaiton links in lib.rs

its-luca:
The edited links in the main lib.rs documentation lead nowhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,11 +47,11 @@ To include the SEV-SNP APIs only:
 
 ### Platform Management
 
-Refer to the [`firmware`] module for more information.
+Refer to the [firmware] module for more information.
 
 ### Guest Management
 
-Refer to the [`launch`] module for more information.
+Refer to the [launch] module for more information.
 
 ### Cryptographic Verification
 
@@ -74,7 +74,7 @@ with the device nodes.
 
 ### Using the C API
 
-Projects in C can take advantage of the C API for the SEV [`launch`] ioctls.
+Projects in C can take advantage of the C API for the SEV [launch] ioctls.
 To install the C API, users can use `cargo-c` with the features they would
 like to produce and install a `pkg-config` file, a static library, a dynamic
 library, and a C header:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,11 +42,11 @@
 //!
 //! ## Platform Management
 //!
-//! Refer to the [`firmware`] module for more information.
+//! Refer to the [firmware] module for more information.
 //!
 //! ## Guest Management
 //!
-//! Refer to the [`launch`] module for more information.
+//! Refer to the [launch] module for more information.
 //!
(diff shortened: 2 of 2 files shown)


tylerfanelli:
Thanks. Can you run `cargo readme` and amend the changes to the commit?

its-luca:
If I understand correctly this would mean that the edited parts would no longer be a link in markdown (tested with the VSCode markdown renderer, not sure if Github would render that differently). Is that ok?

DGonzalezVillal (approved):

tylerfanelli:
I think it should be fine.

tylerfanelli (approved):

tylerfanelli:
@DGonzalezVillal can you review?

tylerfanelli (commented):
So the links would no longer work on GitHub, but they would work on rust documentation?

DGonzalezVillal:
> So the links would no longer work on GitHub, but they would work on rust documentation?

Yeah in his working repo, the links don't work on the GitHub. @tylerfanelli Is there a way keep it working both ways? What's the point of fixing it in one, if it'll break on the other.

tylerfanelli:
We could just remove the [links these modules refer to](https://github.com/virtee/sev/blob/fb28b1dffc5b52aad055cff810cae094cf66c8d9/README.md?plain=1#L84-L85), as they are paths within the github repo. This would remove the links, but the labels `firmware` and `launch` would still be code blocks, indicating that they are source code modules.

DGonzalezVillal:
> We could just remove the [links these modules refer to](https://github.com/virtee/sev/blob/fb28b1dffc5b52aad055cff810cae094cf66c8d9/README.md?plain=1#L84-L85), as they are paths within the github repo. This would remove the links, but the labels `firmware` and `launch` would still be code blocks, indicating that they are source code modules.

And would this have the links working on the rust documentation then?

tylerfanelli:
No, they wouldn't be hyperlinks anymore. They would just be "code blocks" like `this`. Users would still understand that we're referring to a source code module then, which is fine (even though there is no hyperlink).

DGonzalezVillal:
Looking at this link: https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#motivation

Would doing:
`[firmware](sev::firmware)`
work in both secenarios?

tylerfanelli:
Seems like it would work in Rust docs, have you tried in GitHub?

DGonzalezVillal:
So with cargo-readme it won't work. Also cargo-readme has had limited contributions in the past 6 months.

I found this other project that has had recent contributions and you can generate links for both rust.doc and README. Would you be interested in switching to fix this issue?

https://github.com/orium/cargo-rdme

tylerfanelli:
Sure, if it works better for us. Thanks for investigating this! Can you submit a PR with the changes?

PR #171, opened 2024-04-12
Docs: Fixing broken documentation links.

larrydewey:
Utilizing `cargo-rdme` as an alternative to `cargo readme` to fix the broken links between github and crates.io.

Resolves: #169


tylerfanelli (approved):
Thanks! LGTM.

PR #186, opened 2024-05-07
vlek: Adding vlek ioctls

larrydewey:
Adding in ioctls for SNP_VLEK_LOAD to allow platform owners to load VLEK where appropriate.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -174,6 +174,9 @@ pub enum UserApiError {
     /// Uuid parsing errors.
     UuidError(uuid::Error),
 
+    /// VLEK Hashstick errors.
+    HashstickError(HashstickError),
+
     /// Invalid VMPL.
     VmplError,
 
@@ -188,6 +191,7 @@ impl error::Error for UserApiError {
             Self::FirmwareError(firmware_error) => Some(firmware_error),
             Self::UuidError(uuid_error) => Some(uuid_error),
             Self::VmmError(vmm_error) => Some(vmm_error),
+            Self::HashstickError(hashstick_error) => Some(hashstick_error),
             Self::VmplError => None,
             Self::Unknown => None,
         }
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -238,6 +249,35 @@ impl std::convert::From<CertError> for UserApiError {
...
+    /// Certificate length does not match what was specified in the buffer.
+    InvalidLength,
+
+    /// No certificates were provided
+    EmptyCertBuffer,
+
+    /// Unknown Error.
+    UnknownError,
+}
+
```
DGonzalezVillal, line 264: Why create a VLEK error instead of using the existing Cert error?

src/firmware/linux/host/types/snp.rs
```diff
@@ -249,6 +249,71 @@ impl Default for SnpSetConfig {
...
+
+#[cfg(feature = "snp")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(C, packed)]
+/// Wrapped VLEK data.
+pub struct WrappedVlekHashstick<'a> {
+    /// Opaque data provided by AMD Key Distribution Server
+    /// (as described in SEV-SNP Firmware ABI 1.54, SNP_VLEK_LOAD)
+    pub data: &'a [u8], // 432 bytes of data
+}
```
DGonzalezVillal, line 263: So the KDS provides bytes not an actual x509 cert?

src/firmware/linux/host/types/snp.rs
```diff
@@ -249,6 +249,71 @@ impl Default for SnpSetConfig {
...
+
+#[cfg(feature = "snp")]
+impl SnpVlekLoad {
+    /// Creates a new VLEK load instruction from a hashstick.
+    pub fn new(hashstick: &WrappedVlekHashstick) -> Self {
+        hashstick.into()
+    }
+}
+
```
DGonzalezVillal, line 305: So the use scenario for a user is that they provide the raw bytes into a SNPVlekLoad struct?

DGonzalezVillal (commented):
Just a couple of discussion points, but the code looks good to me.

src/error.rs
larrydewey: When the platform owner calls SNP_VLEK_LOAD, they are passing a Wrapped Hashstick, not the x509 certificate itself. The certificate is provided to the guest via the extended guest request workflow.

larrydewey: These errors are separate, as they are not certificate errors per se.

src/firmware/linux/host/types/snp.rs
larrydewey: Correct. It is a wrapped hashstick

src/firmware/linux/host/types/snp.rs
larrydewey: In this case, the user will provide the hashstick blob to the API, and the API will wrap the contents into a `WrappedVlekHashstick`, then that will be wrapped into the `SnpVlekLoad` command structure, which is subsequently passed to the IOCTL. Once the hashstick is loaded into the ASP, the platform owner will provide the certificate to the guest via the extended guest request mechanism.

src/error.rs
larrydewey: Do you think maybe we refer to them has `HashstickErrors`, instead?

DGonzalezVillal (approved):
LGTM

PR #206, opened 2024-07-26
certs: Adding additional functionality.

larrydewey:
- Made X509 within Certificate pub (requires major version bump).
- Added in a bunch of additional support for translating between `Certificate`, `openssl::X509`, `ca::Chain`, and `Chain`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sev"
-version = "3.2.0"
+version = "3.3.0"
 authors = [
     "Nathaniel McCallum <npmccallum@redhat.com>",
     "The VirTEE Project Developers",
--- a/src/certs/snp/ca/mod.rs
+++ b/src/certs/snp/ca/mod.rs
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
+#[cfg(feature = "openssl")]
+use openssl::x509::X509;
 
 use super::*;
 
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sev"
-version = "3.2.0"
+version = "3.1.1"
```
DGonzalezVillal: Change version before merge

tylerfanelli (changes requested):
There's a lot of assumptions we're making about the formatting of the inputs (for instance, an array should be ARK/ASK/VCEK). Documentation for the rustdocs should make this abundantly clear.

Each conversion trait `impl` should probably outline this.

DGonzalezVillal (dismissed):
LGTM!

DGonzalezVillal:
> There's a lot of assumptions we're making about the formatting of the inputs (for instance, an array should be ARK/ASK/VCEK). Documentation for the rustdocs should make this abundantly clear.
> 
> Each conversion trait `impl` should probably outline this.

ASK/ARK/VEK, but yes I agree, it should be made clear for users

larrydewey:
Changes have been made on the latest commit.

DGonzalezVillal (approved):

src/certs/snp/ca/mod.rs
```diff
@@ -29,6 +31,77 @@ impl<'a> Verifiable for &'a Chain {
     }
 }
 
+#[cfg(feature = "openssl")]
+impl From<(X509, X509)> for Chain {
```
tylerfanelli, line 35: I think the ordering of the tuples/arrays/etc should be ARK/ASK rather than ASK/ARK.

tylerfanelli (changes requested):

src/certs/snp/ca/mod.rs
larrydewey: I don't disagree that I would like the root key first, however we need to follow the standard already presented by the AMD Key Distribution Server (KDS) which explicitly returns (ASK, ARK) in PEM format:

https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/57230.pdf#page=17

tylerfanelli (approved):

PR #210, opened 2024-07-31
build: Fixing linting issue for build.rs

larrydewey:
Previously we were not registering the custom flags for rustc to check. Adding these two lines fixes a bunch of linting issues

EDIT: For Context see [rust-check-cfg](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-check-cfg)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,6 +3,11 @@
 fn main() {
     use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


larrydewey:
Just kidding. This is only when using the latest version of Rust. After using `rustup override set 1.70.0` these errors go away.

PR #211, opened 2024-07-31
clippy: Adding clippy::style check

larrydewey:
Adding additional clippy style check to clean up linting errors.


tylerfanelli (approved):

DGonzalezVillal (approved):
-D clippy::style catches some extra style fixes  we should be aware about. The linting errors Larry is fixing appear on later versions of rust (latest), Fixing those does not break 1.70, so I'm good fixing those now to avoid linting errors later.

Issue #189, opened 2024-05-16
Randomness Generation - Guarantee its from SEV Enclave. 


Astewart1510:
Hi there,

I am migrating from Intel SGX with Gramine, where we can request randomness securely sourced from the enclave. I am trying to find a similar feature in AMD SEV.

I noticed that the sev crate uses OpenSSL to generate randomness, which likely leverages the system's RNG. However, does this method guarantee that the randomness is sourced specifically from the SEV enclave?

Could you provide advice or technical guidance on how to generate randomness securely within the SEV enclave?

Thank you!

larrydewey:
For clarification @Astewart1510, are you using SEV/SEV-ES or are you using SEV-SNP?

larrydewey:
> I noticed that the sev crate uses OpenSSL to generate randomness, which likely leverages the system's RNG. However, does this method guarantee that the randomness is sourced specifically from the SEV enclave?

No. Unfortunately using the openssl RNG does not guarantee all randomness generated by `/dev/urandom` (what seeds the RNG for OpenSSL) originated from the enclave itself. The Linux kernel will add sources of randomness, potentially from the VMM/Hypervisor, to be mixed into the random sources. While it is likely to be okay, it does not guarantee a singular origin.

> Could you provide advice or technical guidance on how to generate randomness securely within the SEV enclave?

One of the best sources for randomness would be calling the x86 instruction `RDRAND`, as all CPU instructions are protected by the Guest Hypervisor Communication Block (GHCB), and may be considered trusted. The CPU instruction will generate randomness from the AMD Secure Processor SOC.

Astewart1510:
Thank you! apologies for my delayed response here.

PR #213, opened 2024-08-02
rand: Removing use of openssl rand for crypto

larrydewey:
According to the documentation, openssl's `rand_bytes(...)` function will be seeded with `/dev/urandom`, which is not sound for cryptographic application. We will be switching to using the `rdrand` crate which will utilize the X86_64 microcode call to `RDRAND`, instead. All X86_64 instructions are protected by the Guest Hypervisor Communication Block (GHCB), and may thus be presumed secure for cryptographic practices in the context of confidential compute.

Obviously this introduces a change to the API, requiring a new major release of the crate.

Resolves: #189

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,6 +703,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rdrand"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


tylerfanelli:
Before we do this, we should discuss the [potential drawbacks of RDRAND](https://lwn.net/Articles/961121/).

PR #242, opened 2024-11-06
Adding from PEM stack for CA chain functionality

DGonzalezVillal:
Adding the ability to create a CA chain from a PEM stack. Will be useful when getting the certificates from the KDS, since that is how those certificates are provided. Will also fix the issue of different cert sizes for different generations.


tylerfanelli:
Can you write a test for `cargo test` to ensure its running as expected?

DGonzalezVillal:
@tylerfanelli @larrydewey Can you guys review

tylerfanelli:
Seems there's a failing test with `crypto_nossl`.

DGonzalezVillal:
@tylerfanelli @larrydewey Can you guys review

src/certs/snp/builtin/mod.rs
```diff
@@ -8,4 +8,7 @@ pub mod genoa;
 /// Milan generation.
 pub mod milan;
 
+/// TURIN generation
```
tylerfanelli, line 11:
```suggestion
/// Turin generation.
```

tylerfanelli (approved):
Slight formatting suggestion, but not a blocker.

PR #250, opened 2024-11-19
feature(measurement): allowing measurements under `crypto_nossl` feature

jakubDoka:
Make the `measurement` module compatible with the `crypto_nossl` feature, so `openssl` feature is not required.

### DIFF
--- a/src/certs/snp/ca/mod.rs
+++ b/src/certs/snp/ca/mod.rs
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-
 //! Operations for a Certificate Authority (CA) chain.
 
 #[cfg(feature = "openssl")]
--- a/src/firmware/guest/types/snp.rs
+++ b/src/firmware/guest/types/snp.rs
@@ -197,8 +197,8 @@ impl ReportVariant {
             0 | 1 => return Err(std::io::ErrorKind::Unsupported.into()),
             2 => Self::V2,
             _ => {
-                let family: u8 = bytes[388];
-                if family >= 0xB0 {
+                let family: u8 = bytes[392];
+                if family >= 0x1A {
                     Self::V3Turin
                 } else {
                     Self::V3PreTurin


src/certs/snp/ca/mod.rs
```diff
@@ -1,11 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
+//! Operations for a Certificate Authority (CA) chain.
```
tylerfanelli, line 2:
```suggestion

//! Operations for a Certificate Authority (CA) chain.
```

src/measurement/gctx.rs
```diff
@@ -12,6 +13,14 @@ use crate::{
     measurement::snp::{SnpLaunchDigest, LD_BYTES},
 };
 
+#[cfg(not(feature = "openssl"))]
```
tylerfanelli, line 16: I don't believe this is correct, as it likely wouldn't work when both `openssl` and `crypto_nossl` features are disabled.

```suggestion
#[cfg(all(not(feature = "openssl"), feature = "crypto_nossl")))]
```

src/measurement/sev_hashes.rs
```diff
@@ -18,6 +20,14 @@ use crate::error::*;
 
 type Sha256Hash = [u8; 32];
 
+#[cfg(not(feature = "openssl"))]
```
tylerfanelli, line 23: Similarly here, this wouldn't work when both `openssl` and `crypto_nossl` features are disabled.

```suggestion
#[cfg(all(not(feature = "openssl"), feature = "crypto_nossl")))]
```

tylerfanelli (changes requested):

tylerfanelli:
Also, could you please provide a more in-depth commit title and description, as well as a `Signed-off-by` to the commit?

src/measurement/gctx.rs
jakubDoka: Is it a convention to always mention exhaustive conditions? The parent module already covers this case.

tylerfanelli:
Yea, I guess it was exhaustive and not necessary due to the parent module already having the `any(openssl, crypto_nossl)`, but i guess it doesn't hurt. Please squash the commits and otherwise LGTM.

src/measurement/gctx.rs
```diff
@@ -12,6 +13,14 @@ use crate::{
     measurement::snp::{SnpLaunchDigest, LD_BYTES},
 };
 
+#[cfg(all(not(feature = "openssl"), feature = "crypto_nossl"))]
```
DGonzalezVillal, line 16: The library won't compile if both crypto_nossl and openssl are both passed as features, so you can just gate under crypto_nossl

```suggestion
#[cfg(feature = "crypto_nossl")]
```

src/measurement/sev_hashes.rs
```diff
@@ -18,6 +20,14 @@ use crate::error::*;
 
 type Sha256Hash = [u8; 32];
 
+#[cfg(all(not(feature = "openssl"), feature = "crypto_nossl"))]
```
DGonzalezVillal, line 23: Same thing 
```suggestion
#[cfg(feature = "crypto_nossl")]
```

DGonzalezVillal (commented):
Just nit pick, once you fix the testing, we can make sure it is working as intended and then you should be good!

DGonzalezVillal:
And you still need to sign-off the commit

tests/measurement.rs
```diff
@@ -363,7 +363,7 @@ mod snp_tests {
     }
 }
 
-#[cfg(all(target_os = "linux", feature = "sev"))]
+#[cfg(all(target_os = "linux", feature = "sev", feature = "openssl"))]
```
DGonzalezVillal, line 366: Did you only add support for SNP?

src/measurement/gctx.rs
jakubDoka: this is like this to prevent two sha384 to be in scope if both openssl  and crypto_nossl is enabled

tests/measurement.rs
jakubDoka: Yes I guess the commit title says otherwise, I can add that too

jakubDoka:
https://github.com/virtee/sev/blob/main/src/lib.rs#L89

here I see you are enforcing xor featuring, that can be incredibly annoying if you use this crate indirectly and other features end up enabled out of your control, the rust features should just work together no matter the combination you choose

jakubDoka:
> When the crypto_nossl feature was implemented, it essentially just re-wrote functions that used openssl to use an alternative. To avoid redundancy in code the decision was you had to pick if you were going to use openssl or if you were going to use no_ssl, but you can't compile both at the same time. It makes sense although because why would you want 2 features that do the same thing? It also reduces package dependencies at compile time and avoids other issues. What features are being enabled out of your control?

Let's say you depend on a crate `A` that uses `sev` and since it does not need to use any openssl-only features, the author decides to use crypto_nossl, but then you also import crate `B`, this crate is using some features of the `sev` that require `openssl` thus you end up with both openssl and crypto_nossl being enabled because rust will not compiler the crate for each set of features it's used with, it takes a union of features from all imports. Once you get to this configuration, you will get the compiler error from a dependency and your only saving grace is either, the author of a crate `A` offers you to switch to the `openssl` feature, or not use either of `A` or `B`.

Another scenario where this is annoying is when you have .eg wasm and native crates in your project and since `openssl` does not work in that environment, you are forced to use `crypto_nossl`, then if you still need to use `openssl` in the native crate, `rust-analyzer` gets bricked on the compiler error and you need to the do conditional imports in your cargo toml to pick different features for native and wasm even though the crate is never compiled to native code.

Note: if it's not yet possible to compile this crate to wasm, I have an interest in making that possible, but first things need to not depend on OpenSSL

src/measurement/gctx.rs
tylerfanelli: This was my mistake, apologies @jakubDoka

src/measurement/sev_hashes.rs
tylerfanelli: Again, my mistake. I asked @jakubDoka forgetting that it won't compile with both features.

tylerfanelli:
So you're suggesting that we allow both `openssl` and `crypto_nossl` features concurrently (ie removing the compiler error) in this case? I'm not totally against it, but I would expect that if both are enabled, one of them (IMO `openssl`) would take precedent. That is, if you enable both `crypto_nossl` AND `openssl`, it would essentially be the same as enabling `openssl` only.

jakubDoka:
No, in the case of compiling to the Wasm example, this is only an issue when using a workspace and Rust analyzer. Once you build the Wasm crate, you need to make sure OpenSSL is not enabled, which is the case because you have to compile each architecture separately. It's just that rust-analizer runs `cargo check` for all crates in the workspace

jakubDoka:
@tylerfanelli In any case, I can still apply the changes @DGonzalezVillal is asking for, xor featuring is not that important, I mainly want to move on with this. I have more PRs planned as I need snp measurement and guest side of attestation to be cross-platform (macOS, Windows, WASM if possible)

ghe0:
Hello!

Can you please let us know which modifications are still needed to close this PR?

src/measurement/sev.rs
```diff
@@ -77,7 +84,12 @@ pub fn seves_calc_launch_digest(
         launch_hash.update(vmsa_page.as_slice())
     }
 
-    Ok(launch_hash.finish())
+    #[cfg(feature = "openssl")]
+    {
+        Ok(launch_hash.finish())
+    }
```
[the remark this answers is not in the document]
DGonzalezVillal: @jakubDoka @ghe0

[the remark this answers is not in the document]
jakubDoka: @larrydewey @DGonzalezVillal https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1e66295c50ba28519a05d11bc4f37d60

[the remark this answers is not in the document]
ghe0: @larrydewey @DGonzalezVillal that code works for me. Can you please post the errors you are getting?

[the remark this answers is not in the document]
DGonzalezVillal: @larrydewey It seems to me that it works. Would you prefer defining this as different functions depending on the feature they want to implement?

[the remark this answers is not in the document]
larrydewey: Where this is a one-line thing, I think I am okay with just following the current implementations. I am still researching when this functionality was added, as previously this did not compile.
```suggestion
    #[cfg(feature = "openssl")]
    Ok(launch_hash.finish())
```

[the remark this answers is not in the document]
DGonzalezVillal: @larrydewey would you prefer if it is split into 2 functions instead?

src/measurement/sev.rs
```diff
@@ -110,5 +125,10 @@ pub fn sev_calc_launch_digest(
         launch_hash.update(sev_hashes.as_slice());
     };
 
-    Ok(launch_hash.finish())
+    #[cfg(feature = "openssl")]
+    {
+        Ok(launch_hash.finish())
+    }
```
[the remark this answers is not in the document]
ghe0: @larrydewey did you test this out? Does it work for you now?

PR #252, opened 2024-11-21
Fixed testing to have comprehensive testing for the library

DGonzalezVillal:
Several bugs caught when trying to run cargo test with different feature combinations. Fixed these bugs and made some slight tweaks to be able to run comprehensive testing on the library.

- When running cargo test with dangerous_hw_tests enabled both has_sev and has_sev_guest tests were run leading to a failure. Fixed by removing an unecessary or
- Renamed has_sev to host and renamed has_sev_guest to guest
- ID-BLOCK had to be put behind openssl only
- Changed snp_config test to hit only the IOCTL and not modify the config. This was done to avoid messing up system configs during testing.
- Renamed the launch test to sev_launch test. This allows us to test that launch test individually.
- Added code to create a legacy SEV cert-chain locally. This allows us to test the SEV launch test without the need of an external tool.
- Cached Turin certificates to the SEV builtin certs and the SNP builtin certs.


larrydewey (approved):

tylerfanelli:
The Turin certificate additions that were originally in #242 are also included in this commit. Can you rebase on main and remove the cert additions?

DGonzalezVillal:
Bumped mac-os version in linter

DGonzalezVillal:
@tylerfanelli

src/certs/sev/builtin/turin/ark.cert
tylerfanelli, line 1: Were these not added in #242 ? Edit: I see the SNP builtin certs were added. Are they different?

src/util/cached_chain.rs
```diff
@@ -11,21 +11,31 @@
 //! An entire certificate chain can be created using the `sevctl`
 //! utility.
 
-#[cfg(feature = "sev")]
-use crate::certs::sev::Chain;
+#![cfg(all(feature = "sev", feature = "dangerous_hw_tests"))]
```
tylerfanelli, line 14: What is this for?

tylerfanelli (commented):
Two quick comments, and this can be merged.

tylerfanelli (approved):

PR #253, opened 2024-11-25
Adding deprecation flags for SEV components and bumping to 5.1

DGonzalezVillal:
As discussed, adding deprecation flags for SEV components, so that it can be removed in the next 6.0 release. Bumping to 5.1 so that people can still use 5.0.0 without compilation warnings if they are using SEV.

### DIFF
--- a/src/certs/sev/builtin/mod.rs
+++ b/src/certs/sev/builtin/mod.rs
@@ -4,9 +4,24 @@
 //!
 //! These are primarily offered as a convenience measure to avoid making
 //! HTTP requests to AMD's servers.
-
+#[deprecated(
+    since = "5.0.0",
+    note = "Legacy SEV features will no longer be included/supported in library versions past 5"
+)]
 pub mod genoa;
+#[deprecated(
+    since = "5.0.0",
+    note = "Legacy SEV features will no longer be included/supported in library versions past 5"
+)]
 pub mod milan;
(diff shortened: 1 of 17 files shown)


larrydewey (approved):
LGTM

DGonzalezVillal:
Adding WIP flag, waiting for 5.0 release first

tylerfanelli:
Set to draft PR as well, so this can't be merged.

DGonzalezVillal:
Can be merged now if we want this still

src/certs/sev/builtin/mod.rs
```diff
@@ -4,9 +4,24 @@
 //!
 //! These are primarily offered as a convenience measure to avoid making
 //! HTTP requests to AMD's servers.
-
```
tylerfanelli, line 7: You left a `deprecated` note on this module in `src/certs/mod.rs`. Why are these deprecation notes also needed?

tylerfanelli (changes requested):

larrydewey (approved):

tylerfanelli:
Let's hold off on this until sev-5.0.0 is packaged for distros. Will revisit this at that point.

tylerfanelli:
After further consideration and discussing with my some other stakeholders, I don't think it's the best time to do this.

PR #259, opened 2024-12-04
Adding the KeyInfo field into the attestation report

DGonzalezVillal:
Since version 2 of the attestation report, instead of having just the AUTHOR_KEY_EN fIELD in the attestation report, the report now offers more data around the signature and the key being used in the attestation report. This PR extends that field and makes it accessible in the attestation report.

(This has been in the spec since sept 2023, unrelated to upcoming changes)


larrydewey (approved):
Good catch!

tylerfanelli (approved):

tylerfanelli:
`KeyInfo` is still represented as a `u32`, so this is fine. Thanks for this.

PR #265, opened 2024-12-18
firmware/snp: Update 0x3 platform initialization data

tylerfanelli:
Initially, byte 0x3 of STRUCT_PLATFORM_STATUS only used bit 0 to indicate if the RMP was initialized. Updates were made to include bitfields to indicate if alias detection has completed since the last system reset and there are no aliasing addresses (bit 1), as well as if SEV-TIO is enabled (bit 3).

### DIFF
--- a/src/firmware/host/types/snp.rs
+++ b/src/firmware/host/types/snp.rs
@@ -225,6 +225,24 @@ pub struct Build {
     pub build: u32,
 }
 
+bitflags::bitflags! {
+    /// Various platform initialization configuration data. Byte 0x3 in SEV-SNP's
+    /// STRUCT_PLATFORM_STATUS.
+    #[derive(Default)]
+    pub struct PlatformInit: u8 {
+        /// Indicates if RMP is initialized.
+        const IS_RMP_INIT           = 1 << 0;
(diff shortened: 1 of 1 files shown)


DGonzalezVillal (approved):
lgtm

src/firmware/host/types/snp.rs
```diff
@@ -225,6 +225,21 @@ pub struct Build {
...
+        /// Indicates if RMP is initialized.
+        const IS_RMP_INIT           = 1 << 0;
+        /// Indicates that alias detection has completed since the last system reset
+        /// and there are no aliasing addresses. Resets to 0.
+        const ALIAS_CHECK_COMPLETE  = 1 << 1;
+        /// Indicates TIO is enabled. Present if SevTio feature bit is set.
+        const IS_TIO_EN             = 1 << 3;
+    }
+}
+
```
DGonzalezVillal, line 242: Should we maybe add in what FW version (1.55.38 for Genoa and 1.55.22 for Milan) this was added?

tylerfanelli: Sure, I've added a comment indicating this.

PR #266, opened 2024-12-19
Adding AttestationReport Verification with Certificate only

DGonzalezVillal:
We currently only offer report verification with a full chain being provided. Although it is safest to provide a whole chain to verify the whole root of trust, there are cases people might want to divide the process into individual steps. This allows users to verify the report with what should be an already verified versioned endorsement key.

I also renamed the VCEK to VEK since we no offer support for both the VLEK and VCEK.

### DIFF
--- a/src/firmware/guest/types/snp.rs
+++ b/src/firmware/guest/types/snp.rs
@@ -3,7 +3,7 @@
 use crate::{certs::snp::ecdsa::Signature, firmware::host::TcbVersion, util::hexdump};
 
 #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
-use crate::certs::snp::{Chain, Verifiable};
+use crate::certs::snp::{Certificate, Chain, Verifiable};
 
 use std::fmt::Display;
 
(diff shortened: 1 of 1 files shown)


src/firmware/guest/types/snp.rs
```diff
@@ -345,13 +377,57 @@ impl Verifiable for (&Chain, &AttestationReport) {
 impl Verifiable for (&Chain, &AttestationReport) {
     type Output = ();
 
+    fn verify(self) -> io::Result<Self::Output> {
+        // According to Chapter 3 of the [Versioned Chip Endorsement Key (VCEK) Certificate and
+        // KDS Interface Specification][spec] and the [Versioned Loaded Endorsement Key Certificate Definition][spec2],
+        // both Versioned Endorsement Key certificates certify an ECDSA public key on curve P-384,
+        // and the signature hash algorithm is sha384.
+        // [spec]: https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/57230.pdf
```
tylerfanelli, line 385: I don't think these two links are necessary. Rather, you can mention "see specifications _X_ and _Y_".

tylerfanelli (changes requested):

DGonzalezVillal:
@tylerfanelli @larrydewey Re-review please

tylerfanelli (approved):

PR #268, opened 2024-12-20
Attestation Report versioning Update

DGonzalezVillal:
In spec 1.56 of the SEV firmware a new version of the attestation report was introduced.

Here we are introducing a way to version the attestation report that keeps security and backwards compatibility.

The main AttestationReport is now an enum that will contain the different versions of the attestation report. There are 2 new structs for the Attestation Report, one for each version. There is a new trait called Attestable that all the attestation reports will implement, this will allow users to attest their report regardless of the version.

The ReportRsp will now contain raw bytes, rather than the Attestation Report Strucutre. The AttestationReport Enum has a TryFrom bytes that will return the appropriate attestation report version according to the first 4 bytes of the raw data.

Structs consumed by the attestation report that now have new fields depending on the version, are now also versioned, and each report will consume the appropriate version of that struct (look at PlatInfo).


DGonzalezVillal:
This is a possible solution to the attestation report versioning issue. I would appreciate thoughts and concerns around this approach.

src/firmware/guest/types/snp.rs
```diff
@@ -309,23 +367,263 @@ Launch TCB:
...
+            self.committed_minor,
+            self.committed_major,
+            self.launch_tcb,
+            self.signature
+        )
+    }
+}
+
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+impl Attestable for AttestationReportV3 {
```
larrydewey, line 596: You should implement these as default behavior in the trait, and then override only when different behavior is required.

larrydewey (changes requested):
One nit

larrydewey:
CC: @fitzthum

src/error.rs
```diff
@@ -699,6 +712,33 @@ impl std::fmt::Display for CertError {
 
 impl error::Error for CertError {}
 
+#[derive(Debug)]
+/// Errors which may be encountered when handling attestation reports
+pub enum AttestationError {
```
tylerfanelli, line 717:
```suggestion
pub enum AttestationReportError {
```

Perhaps this is a bit more descriptive?

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +113,76 @@ bitfield! {
...
+/// The firmware signs the attestation report with its VEK (VCEK or VLEK). The firmware uses the system wide
 /// ReportedTcb value as the TCB version to derive the VCEK. This value is set by the hypervisor.
+/// The VLEK is generated externally and has to be loaded into the machine.
 ///
 /// The firmware guarantees that the ReportedTcb value is never greater than the installed TCB
 /// version
+///
+/// Since the release of the 1.56 ABI, the Attestation Report was bumped from version 2 to 3.
+/// Due to content differences, both versions are kept separately in order to provide backwards compatibility and most reliable security.
+pub enum AttestationReport {
```
tylerfanelli, line 125:
```suggestion
pub enum AttestationReport {
```

I think this is probably the best way to do things given that we can't reasonably expect a user to know which version they're getting.

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +113,76 @@ bitfield! {
...
+        match self {
+            AttestationReport::V2(report) => report.measurable_bytes(),
+            AttestationReport::V3(report) => report.measurable_bytes(),
+        }
+    }
+}
+
+/// Trait shared between attestation reports to be able to verify them against the VEK.
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+trait Attestable {
```
tylerfanelli, line 175: Can you define this trait before you `impl` it for `AttestationReport`? Just for readability.

src/firmware/linux/guest/types.rs
```diff
@@ -156,6 +156,8 @@ impl ReportReq {
     }
 }
 
+const REPORT_SIZE: usize = std::mem::size_of::<AttestationReportV2>();
```
tylerfanelli, line 159: Are all reports expected to be the same size?

tylerfanelli (changes requested):
Some small changes, but overall I think this LGTM and the dealing with different versions of attestation reports is handled nicely.

fitzthum (commented):
Looks fine. It's important to support different report versions and generally to decouple the version of the crate from the version of the platform.

This will add complexity for users of the crate. It's really handy that both reports can be attested in the same way, but retrieving values from the reports will require some conditional logic (potentially even if it's a value that both report types support). Keep in mind that we are likiely to have more report versions as time goes on.

This complexity is sort of inevitable, but there might be ways to mitigate it. For instance, you might introduce another trait with some helper functions that expose values that are common to the report. The reports are mostly the same after all. Maybe there is more that can be shared.

As an aside, there isn't much documentation or many examples for this crate. It's beyond the scope of this PR, but if there some more examples about usage, it might help to understand the scope of this kind of change.

Will this lead to a major release?

tylerfanelli:
I wonder if it's moreso worthwhile to define AttestationReport as a **trait** rather than an **enum**, and have all common values of an attestation report be addressable by trait methods rather than struct members. This may alleviate some of the complexity with dealing with different attestation report versions.

tylerfanelli:
My above comment is expanding a bit on what @fitzthum puts succinctly here:

> 
> This complexity is sort of inevitable, but there might be ways to mitigate it. For instance, you might introduce another trait with some helper functions that expose values that are common to the report. The reports are mostly the same after all. Maybe there is more that can be shared.

tylerfanelli:
Lets hold off on this major release until all of those changes (remove SEV module, etc) are in. This firmware isn't generally available yet, correct?

src/firmware/linux/guest/types.rs
DGonzalezVillal: So far the different versions should be the same size. We do not expect to change any time soon.

DGonzalezVillal:
@tylerfanelli @fitzthum @larrydewey 

Hey everyone,

I spent some time exploring how we could refactor the changes so that different Attestation Report versions implement an `AttestationReport` trait rather than using an `enum.` However, this approach introduced more complexity, particularly in the API functions that call the SNP_GET_REPORT IOCTL.

Here’s what the trait might look like:

The trait looks something like this: 
```rust
pub trait AttestationReport: Display + Attestable {   
    fn version(&self) -> u32;
}
````
The idea was to define getter functions for the relevant fields and initially implement the get_report function with a return type like this:
```rust
 pub fn get_report(
        &mut self,
        message_version: Option<u32>,
        data: Option<[u8; 64]>,
        vmpl: Option<u32>,
    ) -> Result<impl AttestationReport, UserApiError>
````

However, while building this out, I realized it wouldn’t work because the return type isn’t consistent. The PSP generates different report structures depending on the version, so we can’t return the same concrete type every time. Instead, the return type needs to be a dynamically boxed AttestationReport.

Here’s the resulting implementation of the get_report function:

```rust
pub fn get_report(
        &mut self,
        message_version: Option<u32>,
        data: Option<[u8; 64]>,
        vmpl: Option<u32>,
    ) -> Result<Box<dyn AttestationReport>, UserApiError> {
        let mut input = ReportReq::new(data, vmpl)?;
        let mut response = ReportRsp::default();

        let mut request: GuestRequest<ReportReq, ReportRsp> =
            GuestRequest::new(message_version, &mut input, &mut response);

        SNP_GET_REPORT
            .ioctl(&mut self.0, &mut request)
            .map_err(|_| map_fw_err(request.fw_err.into()))?;

        // Make sure response status is successful
        if response.status != 0 {
            Err(FirmwareError::from(response.status))?
        }

        let raw_report = response.report.as_slice();

        let version = u32::from_le_bytes([raw_report[0], raw_report[1], raw_report[2], raw_report[3]]);

        // Return the appropriate report version
        match version {
            2 => {
                let report_v2: AttestationReportV2 = raw_report.as_slice().try_into()?;
                Ok(Box::new(report_v2))
            }
            3 => {
                let report_v3: AttestationReportV3 = raw_report.as_slice().try_into()?;
                Ok(Box::new(report_v3))
            }
            _ => Err(AttestationReportError::UnsupportedReportVersion(version))?,
        }

    }
```
To enable dynamic dispatch for AttestationReport, the trait must be unsized, which introduces limitations. For example, we can’t use traits like TryFrom directly on it.

While this approach isn’t inherently unworkable, it adds significant complexity. It can make the code harder for users to understand and maintain. For instance, extended report requests would have a return type like this:

```rust
pub fn get_ext_report(
        &mut self,
        message_version: Option<u32>,
        data: Option<[u8; 64]>,
        vmpl: Option<u32>,
    ) -> Result<(Box<dyn AttestationReport>, Option<Vec<CertTableEntry>>), UserApiError>
```

This already triggers a `very complex type used` Clippy warning.

Given these trade-offs, I still think using an enum is the best approach for handling different versions. We can introduce a separate trait for accessing report fields, as @fitzthum previously suggested, but the API itself should manage version differences using an enum.

Before I proceed further with this approach, I’d like to hear your thoughts!

DGonzalezVillal:
> Lets hold off on this major release until all of those changes (remove SEV module, etc) are in. The updated firmware isn't generally available yet, correct?

This firmware has already been slowly rolled out to customers already. It is still not out for the general public, but people already have access to it.

DGonzalezVillal:
Couple of  last thoughts on the dynamic vs enum approach.

Using an enum ensures that the returned value is one of the predefined versions of the AttestationReport. While you may not know the specific version at compile time, you can be confident it will match one of the defined enum variants. For example:

```rust
trait AttestationReport {
    fn version(&self) -> u32;
    fn abi_major(&self) -> u8;
    fn v3_unique(&self) -> Option<u8>;
}

/// Trait allows anyone to implement their own structure or functionality, as it is a Trait-only requirement.
struct BogusReport {
    version: u32,
    abi_major: u8,
    v3_unique: u8
}

```
Using an enum ensures that the returned value is one of the predefined versions of the AttestationReport. While you may not know the specific version at compile time, you can be confident it will match one of the defined enum variants. For example:
Enum:
```rust
// Example structures
struct AttestationReportV2 { version: u8}
struct AttestationReportV3 { version: u8, v3_field: u8}

// Still using Attestable trait
trait Attestable {
    fn attest(&self) -> bool;
}

// Example enum
enum AttestationReport {
    V2(AttestationReportV2),
    V3(AttestationReportV3)
}

// Implementing enum
impl Attestable for AttestationReport {
    fn attest(&self) -> bool {
        match self {
            Self::V2(_) => true,
            Self::V3(_) => false
        }
    }
}

// Implementing Getter functions for enum (no need to unwrap for common fields)
impl Test {
    fn version(&self) -> u8 {
        match self {
            Self::V2(report) => report.version,
            Self::V3(report) => report.version
        }
    fn additional_field(&self) -> Option<u8> {
        match self {
            Self::V2(_) => None,
            Self::V3(report) => Some(report.v3_field)
        }
    }
}
```
Trait:
```rust
trait AttestationReport {
    fn version(&self) -> u32;
    fn abi_major(&self) -> u8;
    fn v3_unique(&self) -> Option<u8>;
}

struct AttestationReportV2 {
    version: u32,
    abi_major: u8
}

impl AttestationReport for AttestationReportV2 {
    fn version(&self) -> u32 {
        self.version
    }
    fn abi_major(&self) -> u8 {
        self.abi_major
    }
    fn v3_unique(&self) -> Option<u8> {
        None
    }
}

struct AttestationReportV3 {
    version: u32,
    abi_major: u8,
    v3_unique: u8
}
impl AttestationReport for AttestationReportV3 {
    fn version(&self) -> u32 {
        self.version
    }
    fn abi_major(&self) -> u8 {
        self.abi_major
    }
    fn v3_unique(&self) -> Option<u8> {
        Some(self.v3_unique)
    }
}
```
As you can see, with traits, repetitive code is required for each implementation of the AttestationReport trait. Enums, on the other hand, allow centralized logic for common operations, reducing boilerplate.

DGonzalezVillal:
@tylerfanelli

fitzthum:
@DGonzalezVillal I think you probably will want to have a struct for each report type. It seems like a lot of messing around without that and users probably will want the ability to just deserialize a report of known version and access the fields.

I think the common layer, either a trait or another struct, would live on top of that and give users who don't know/care about the version a way to access the fields.

DGonzalezVillal:
@larrydewey @tylerfanelli @fitzthum 

Hey guys, I made some changes and addressed some of the comments. The AttestationReport Enum now works as an interface that the users can use to reach any of the fields. This way they do not have to manually unwrap the report. They can Also use the enum to display the report. 

If the field that the user is trying to access on the report is not available in the report version they provided, then an error will be raised. I don't know if you would rather this behavior, or instead just return a None value.

Let me know if you have any other comments.

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +128,300 @@ bitfield! {
...
+        }
+    }
+}
+
+impl AttestationReport {
+    /// Get Attestation Report Version
+    pub fn version(&self) -> u32 {
+        match self {
+            Self::V2(report) => report.version,
+            Self::V3(report) => report.version,
```
larrydewey, line 206: You can simplify this into a one-liner which saves typing.
```suggestion
            Self::V2(report) | Self::V3(report)  => report.version,
```

larrydewey (changes requested):

src/firmware/guest/types/snp.rs
DGonzalezVillal: It won't allow you to do this since both reports are not the same type. `Self::V2(report)` is of type `AttestationReportV2` and `Self::V3(report)` is of type `AttestationReportV3`

DGonzalezVillal: https://www.rustexplorer.com/b#LyoKW2RlcGVuZGVuY2llc10KCiovCgpzdHJ1Y3QgQSB7dmVyc2lvbjogdTMyfQpzdHJ1Y3QgQiB7dmVyc2lvbjogdTMyfQoKZW51bSBMZXR0ZXJzIHsKICAgIEEoQSksCiAgICBCKEIpCn0KCmltcGwgTGV0dGVycyB7CiAgICBmbiB2ZXJzaW9uKCZzZWxmKSAtPiB1MzIgewogICAgICAgIG1hdGNoIHNlbGYgewogICAgICAgICAgICBTZWxmOjpBKGxldHRlcikgfCBTZWxmOjpCKGxldHRlcikgPT4gbGV0dGVyLnZlcnNpb24KICAgICAgICB9CiAgICB9Cn0KZm4gbWFpbigpIHsKICAgIHByaW50bG4hKCIiKTsKfQ==

larrydewey: You are totally right. My mistake

larrydewey (approved):

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +128,300 @@ bitfield! {
...
+                let measurable_bytes: &[u8] = &bincode::serialize(report).map_err(|e| {
+                    Error::new(
+                        ErrorKind::Other,
+                        format!("Unable to serialize bytes: {}", e),
+                    )
+                })?;
+                Ok(measurable_bytes[..0x2a0].to_vec())
+            }
+        }
+    }
```
tylerfanelli, line 192: Perhaps `Attestable` can be `impl` for the underlying report versions as well?
```suggestion
impl Attestable for AttestationReport {
    fn measurable_bytes(&self) -> io::Result<Vec<u8>> {
        //Return measurable bytes for the report
        match self {
           Self::V2(v2) => v2.measurable_bytes()?
            Self::V3(v3) => v3.measurable_bytes()?
        }
    }
```

fitzthum (approved):
Looks good. One note

src/firmware/guest/mod.rs
```diff
@@ -91,7 +92,7 @@ impl Firmware {
         message_version: Option<u32>,
         data: Option<[u8; 64]>,
         vmpl: Option<u32>,
-    ) -> Result<AttestationReport, UserApiError> {
+    ) -> Result<impl Attestable, UserApiError> {
```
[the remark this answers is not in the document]
DGonzalezVillal: Forgot to change it after testing a different return types. Will return AttestationReport.

src/firmware/guest/types/snp.rs
```diff
@@ -103,6 +98,22 @@ bitfield! {
     pub get_tcb_version, set_tcb_version: 5, 5;
 }
 
+/// Trait shared between attestation reports to be able to verify them against the VEK.
+pub trait Attestable: Serialize {
```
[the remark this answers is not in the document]
DGonzalezVillal: Like a pub(crate)?

The reason (honestly a lazy approach) it was made public was that the trait implementations are not being used in the crate when the ssl features are not enabled (attestable is only used in verify functions). Whenever openssl or cyrpto_nossl are not enabled, a dead code warning is raised by attestable. Making it public removes the warning. Instead, since we want it to be a crate only trait, I'll gate it behind the cfg options for ssl, and make it a pub(crate).

src/firmware/guest/types/snp.rs
DGonzalezVillal: Good catch!

src/firmware/guest/types/snp.rs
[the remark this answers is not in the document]
tylerfanelli: > Like a pub(crate)?
> 
> The reason (honestly a lazy approach) it was made public was that the trait implementations are not being used in the crate when the ssl features are not enabled (attestable is only used in verify functions). Whenever openssl or cyrpto_nossl are not enabled, a dead code warning is raised by attestable. Making it public removes the warning. Instead, since we want it to be a crate only trait, I'll gate it behind the cfg options for ssl, and make it a pub(crate).

Just to clarify, I would like the trait to be **usable** (i.e. a user of the crate can call `report.measurable_bytes()` just fine) yet not **implementable** (i.e. a user outside of the crate can write `impl Attestable for _____`). I wasn't sure if this was even possible but @larrydewey mentioned it was.

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +128,300 @@ bitfield! {
...
+        }
+    }
+
+    /// Get Platform Info
+    pub fn plat_info(&self) -> PlatformInfoVersion {
+        match self {
+            Self::V2(report) => PlatformInfoVersion::V1(report.plat_info),
+            Self::V3(report) => PlatformInfoVersion::V2(report.plat_info),
+        }
+    }
```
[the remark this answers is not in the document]
tylerfanelli: Why?

src/firmware/guest/types/snp.rs
```diff
@@ -117,14 +128,300 @@ bitfield! {
...
+                        format!("Unable to serialize bytes: {}", e),
+                    )
+                })?;
+                Ok(measurable_bytes[..0x2a0].to_vec())
+            }
+        }
+    }
+    fn signature(&self) -> &Signature {
+        match self {
+            Self::V2(report) => &report.signature,
```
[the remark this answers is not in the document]
DGonzalezVillal: Code ends up being pretty much the same, but I'll change it here too.

src/firmware/guest/types/snp.rs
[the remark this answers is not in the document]
DGonzalezVillal: I see the concern, we might want to be able to reach an actual value instead of another enum to unwrap.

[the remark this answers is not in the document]
fitzthum: Yeah slightly nicer for the user if they can just get a value rather than getting different report types that they have to parse out.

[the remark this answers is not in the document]
DGonzalezVillal: Instead of creating the get functions to get the values here, I created getter functions for the PlatfomInfo enum. Similar to what we are doing here in the AttestationReport, but for that enum. That way they can use that enum directly to get values.

src/firmware/guest/types/snp.rs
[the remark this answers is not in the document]
tylerfanelli: Yea, for this one there's not much change.

src/firmware/guest/types/snp.rs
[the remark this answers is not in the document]
DGonzalezVillal: I see what you mean now. I will look into it, we def do not want pub(crate)

tylerfanelli:
Is this ready for another review? Still has a `WIP` in the PR title.

DGonzalezVillal:
@larrydewey @tylerfanelli @fitzthum 

Please re-review, while doing some testing on the library update, we found a bug around serialization and deserialization and so  I've been playing around on the best way to solve it. This commit (https://github.com/virtee/sev/pull/268/commits/7b5107637a52db20ed0846bf01aae4416c642d4f) explains the issue and solution in depth. Please let me know if you have more questions or comments!

DGonzalezVillal:
Fixed merging issues @tylerfanelli @larrydewey  Let me know your thoughts

larrydewey (approved):

PR #278, opened 2025-02-13
RFC: Switching to custom parser for binary.

larrydewey:
- Refactoring some misplaced code
- Adding custom parsing functionality
- Implemented new and updated existing unit tests to match new functionality
- Implemented parsing library in place of bincode where possible (eventually will completely remove it)


tylerfanelli:
What exactly was wrong with the original binary parsing done? This introduces a few breaking changes that will require a major update.

larrydewey:
Bincode does not do an actual binary deserialization of bytes. It uses a custom format. Serde also relies on binary metadata to determine which instance of an enum something might be. This implementation removes the dependency on bincode to identify attestation report information. It will be a fairly trivial change to adapt this work to the versioning issue, as well. Yes, this will definitely introduce a significant breaking change, but it will align us with a path moving forward for various version change needs.

src/measurement/vmsa.rs
```diff
@@ -206,7 +204,7 @@ struct SevEsSaveArea {
     cpl: u8,
     reserved_0xcc: [u8; 4],
```
DGonzalezVillal, line 205: You meant to leave the once here as std arrays? I think in other structures you changed arrays to Arrays.

DGonzalezVillal (commented):
Quick comments and q2uestions, nothing too mayor

src/firmware/guest/types/snp.rs
```diff
@@ -457,193 +467,258 @@ pub struct AttestationReportV2 {
...
-            current_major: Default::default(),
-            _reserved_2: Default::default(),
-            committed_build: Default::default(),
-            committed_minor: Default::default(),
-            committed_major: Default::default(),
-            _reserved_3: Default::default(),
-            launch_tcb: Default::default(),
-            _reserved_4: [0; 168],
-            signature: Default::default(),
+impl AttestationReport {
```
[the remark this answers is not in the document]
larrydewey: Nope. Just messed it up. Fixed in next commit.

src/firmware/guest/types/snp.rs
```diff
@@ -457,193 +467,258 @@ pub struct AttestationReportV2 {
...
+        handle.write_bytes::<_, 0>(self.image_id())?;
+        handle.write_bytes::<_, 0>(self.vmpl())?;
+        handle.write_bytes::<_, 0>(self.sig_algo())?;
+        handle.write_bytes::<_, 0>(self.current_tcb())?;
+
+        match self {
+            Self::V2(report) => handle.write_bytes::<_, 0>(report.plat_info)?,
+            Self::V3(report) => handle.write_bytes::<_, 0>(report.plat_info)?,
         }
+
```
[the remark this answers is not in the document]
larrydewey: It is a type-specific issue.

src/measurement/vmsa.rs
larrydewey: standard arrays do not implement Display. Array<T, const U: size> does.

DGonzalezVillal (dismissed):
lgtm

DGonzalezVillal (dismissed):
lgtm

src/firmware/guest/mod.rs
```diff
@@ -108,9 +108,7 @@ impl Firmware {
             Err(FirmwareError::from(response.status))?
         }
 
-        let raw_report = response.report.as_array();
-
-        Ok(AttestationReport::try_from(raw_report.as_slice())?)
+        Ok(response.report.to_vec())
```
DGonzalezVillal, line 111: Documentation above will probably have to be changed since we are returning bytes instead of structure.

src/firmware/guest/types/snp.rs
```diff
@@ -643,238 +215,391 @@ impl TryFrom<&[u8]> for AttestationReportV2 {
...
+            },
+        })
+    }
+
+    /// Writes the Attestation Report back into the ASP binary format.
+    pub fn write_bytes(self, mut handle: impl Write) -> Result<(), std::io::Error> {
+        // Determine the variant based on version and CPUID step
+        let variant = if self.version == 2 {
+            ReportVariant::V2
+        } else if self.version == 3 || (self.version == 4 && self.cpuid_step.unwrap_or(0) < 0xB0) {
```
DGonzalezVillal, line 388:
```suggestion
        } else if self.version == 3 || (self.version == 4 && self.cpuid_fam_id.unwrap_or(0) < 0xB0) {
```

Should you be checking family here?

src/firmware/guest/types/snp.rs
```diff
@@ -643,238 +215,391 @@ impl TryFrom<&[u8]> for AttestationReportV2 {
...
+            ReportVariant::V3 | ReportVariant::V4PreTurin => AttestationReport {
+                version: stepper.parse_bytes::<_, 0>()?,
+                guest_svn: stepper.parse_bytes::<_, 0>()?,
+                policy: stepper.parse_bytes::<_, 0>()?,
+                family_id: stepper.parse_bytes::<_, 0>()?,
+                image_id: stepper.parse_bytes::<_, 0>()?,
+                vmpl: stepper.parse_bytes::<_, 0>()?,
+                sig_algo: stepper.parse_bytes::<_, 0>()?,
+                current_tcb: TcbVersion::from_legacy_bytes(&stepper.parse_bytes::<[u8; 8], 0>()?),
+                plat_info: stepper.parse_bytes::<_, 0>()?,
```
DGonzalezVillal, line 331: How are you handling the plat info differences between V2 and V3 +?

src/firmware/guest/types/snp.rs
```diff
@@ -1195,8 +842,8 @@ bitfield! {
     /// Bit 5 indicates that alias detection has completed since the last system reset and there are no aliasing addresses. Resets to 0.
     /// Bits 5-63 are reserved.
     #[repr(C)]
-    #[derive(Default, Deserialize, Clone, Copy, Serialize, PartialEq, Eq, PartialOrd, Ord)]
-    pub struct PlatformInfoV2(u64);
+    #[derive(Deserialize, Clone, Copy, Serialize, PartialEq, Eq, PartialOrd, Ord)]
+    pub struct PlatformInfo(u64);
```
DGonzalezVillal, line 846: Reminder that PlatformInfo Alias check complete is different in V2 and V3+

src/firmware/host/types/snp.rs
```diff
@@ -340,48 +342,101 @@ impl TryFrom<FFI::types::SnpSetConfig> for Config {
...
+            self.tee,
+            self.snp,
+            0,
+            0,
+            0,
+            self.microcode,
+        ]
+    }
+}
+
```
DGonzalezVillal, line 411: So we will keep one display format and fill it out according to the TCB that you should be getting?

src/firmware/guest/mod.rs
larrydewey: Good point. Updated in next commit.

src/firmware/guest/types/snp.rs
larrydewey: Should report default of `false`, so I think we are okay here. We will have to keep an eye on it.

PR #287, opened 2025-03-12
Cfg sev fix

larrydewey:
- Moved `parser` module into utils instead of firmware.
- sev only
- sev + openssl only
- sev + crypto_nossl only
- snp only
- snp + openssl only
- snp + crypto_nossl only


DGonzalezVillal (approved):
lgtm

PR #293, opened 2025-03-26
Fix legacy attestation report validation

rouming:
Hi folks,

For a couple of weeks, I've been investigating AMD SEV technology and experimenting with the `sev/sevctl` tool. I noticed that legacy report validation was only drafted (rewritten from the deprecated `sev-tool`) but actually does not work as expected. I've fixed that and tried to mimic the spirit of the `sevctl` tool as much as possible. This is my first contribution in Rust, so your feedback is highly appreciated.

What has been done in this PR:

* Corrected signature verification to use the SHA256 digest instead of raw data.
* Fixed signature size in `LegacyAttestationReport` by defining it as an `Array<u8, 144>` instead of an `EcdsaSignature` structure, which is 512 bytes.
* Fixed digest size in `LegacyAttestationReport`, ensuring `launch_digest` uses `DIGEST_SIZE` instead of `POLICY_SIZE`.

There should be another PR for `sevctl` itself coming soon, I'll provide a link to that PR once I push it.
Update: the `sevctl` part https://github.com/virtee/sevctl/pull/205

### DIFF
--- a/src/certs/sev/sev/cert/v1/sig/ecdsa.rs
+++ b/src/certs/sev/sev/cert/v1/sig/ecdsa.rs
@@ -75,6 +75,19 @@ impl TryFrom<&[u8]> for Signature {
 }
 
 #[cfg(feature = "openssl")]
+impl TryFrom<&Array<u8, 144>> for ecdsa::EcdsaSig {
+    type Error = Error;
+
+    #[inline]
+    fn try_from(value: &Array<u8, 144>) -> Result<Self> {
+        let arr: &[u8] = value.as_ref();
+        let r = bn::BigNum::from_le(&arr[..SIG_PIECE_SIZE])?;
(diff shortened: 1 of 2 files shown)


DGonzalezVillal (commented):
It looks correct to me, did you have any issues @larrydewey

larrydewey (approved):

rouming:
Please let me know, if there is anything to address from my side. Also I would really appreciate, if you review the `sevctl` part https://github.com/virtee/sevctl/pull/205.

DGonzalezVillal:
@rouming Could you squash into one commit, besides that it looks good, we can merge right after.

rouming:
@DGonzalezVillal @tylerfanelli squashed, please take a look.

tylerfanelli (approved):
LGTM

PR #298, opened 2025-04-08
Adding support for Turin V2 Reports

larrydewey:
Although exceptionally rare, adding support for Turin V2 Reports. Also checking to make sure mask_chip_id is handled appropriately, as well. Should be a patch release.

### DIFF
--- a/src/firmware/guest/types/snp.rs
+++ b/src/firmware/guest/types/snp.rs
@@ -282,42 +282,122 @@ pub struct AttestationReport {
 }
 
 impl AttestationReport {
+    #[inline(always)]
+    /// This will never panic because we are enforcing the length of
+    /// the slice before checking.
+    fn chip_id_is_turin_like(bytes: &[u8]) -> bool {
+        // Chip ID -> 0x1A0-0x1E0
+        // Last 8 bytes of CHIP_ID are zero, then it is Turin Like.
+        bytes == [0; 56]
+    }
+
+    #[inline(always)]
+    /// Check if MASK_CHIP_ID is set before checking chip_id_is_turin_like
+    fn mask_chip_id_en(bytes: &[u8]) -> bool {
+        let mut byte_array: [u8; 4] = [0; 4];
+        byte_array.copy_from_slice(bytes);
+        u32::from_bytes(byte_array) & 0b10 != 0
+    }
+
     /// Attempts to parse an AttestationReport structure from raw bytes.
     pub fn from_bytes(mut bytes: &[u8]) -> Result<Self, std::io::Error> {
         let variant = ReportVariant::from_bytes(bytes)?;
 
         let stepper: &mut &[u8] = &mut bytes;
 
+        if stepper.len() != 1184 {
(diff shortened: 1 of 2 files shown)


src/firmware/guest/types/snp.rs
```diff
@@ -282,42 +282,115 @@ pub struct AttestationReport {
 }
 
 impl AttestationReport {
+    #[inline(always)]
+    /// This will never panic because we are enforcing the lenght of
```
tylerfanelli, line 286:
```suggestion
    /// This will never panic because we are enforcing the length of
```

tylerfanelli (changes requested):
One small nit.

DGonzalezVillal (approved):
LGTM

DGonzalezVillal (approved):

src/firmware/guest/types/snp.rs
```diff
@@ -282,42 +282,122 @@ pub struct AttestationReport {
...
+    /// the slice before checking.
+    fn chip_id_is_turin_like(bytes: &[u8]) -> bool {
+        // Chip ID -> 0x1A0-0x1E0
+        // Last 8 bytes of CHIP_ID are zero, then it is Turin Like.
+        bytes == [0; 56]
+    }
+
+    #[inline(always)]
+    /// Check if MASK_CHIP_ID is set before checking chip_id_is_turin_like
+    fn mask_chip_id_en(bytes: &[u8]) -> bool {
```
tylerfanelli, line 296: The `MASK_CHIP_ID` is never checked in this method? It only checks if the first 4 bytes of some array?

tylerfanelli (approved):
One nit, but overall LGTM.

src/firmware/guest/types/snp.rs
larrydewey: It's sole purpose is to check the second bit of a 32 bit value. Where the method is private, and we are specifically calling it by passing the bytes from the offset specified in the SEV-SNP ABI, it is why I named it and commented as such.

larrydewey: Would you prefer the following, instead:

```rust
fn mask_chip_id_en(bytes: &[u8]) -> bool {
    let mut byte_array: [u8; 4] = [0; 4];
    byte_array.copy_from_slice(bytes);
    KeyInfo(u32::from_bytes(byte_array)).mask_chip_key()
}
 ```

larrydewey: NVM, we can't use the `KeyInfo` fields from the AttestationReport, as they do not include mask_chip_id, but mask_chip_key, and that doesn't indicate if mask_chip_id is enabled or not. Removed that check.

tylerfanelli (approved):

DGonzalezVillal (approved):
lgtm

DGonzalezVillal (approved):

DGonzalezVillal:
Larry just rebased and pushed forced, no need to re review

PR #300, opened 2025-04-16
parser: More friendly ReadExt and WriteExt

larrydewey:
With the previous implementation, it was not a very clean implementation. It worked, but it was full of required type declarations for the `parse_bytes` method. This implements `skip_bytes` separately, and helps segment the process so you don't need to provide it every single time.


DGonzalezVillal (approved):
This LGTM. It really improves readability and coding, so I do appreciate it.

tylerfanelli (approved):

PR #302, opened 2025-04-21
Platform Status Fix

ajcaldelas:
Adding logic to parse bytes from kernel for different tcb versions.

### DIFF
--- a/src/firmware/host/mod.rs
+++ b/src/firmware/host/mod.rs
@@ -4,6 +4,9 @@
 mod types;
 pub use types::*;
 
+#[cfg(feature = "snp")]
+use crate::Generation;
+
 #[cfg(target_os = "linux")]
 use super::linux::host::{ioctl::*, types::GetId};
 
@@ -37,6 +40,9 @@ use std::convert::TryInto;
 #[cfg(feature = "snp")]
 #[cfg(target_os = "linux")]
 use super::linux::host::types::SnpCommit;
+
+#[cfg(all(target_os = "linux", feature = "snp"))]
+use super::linux::host::types::SnpPlatformStatus as FFISnpPlatformStatus;
 
 /// The CPU-unique identifier for the platform.
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -209,14 +215,23 @@ impl Firmware {
     /// ```
     #[cfg(feature = "snp")]
     pub fn snp_platform_status(&mut self) -> Result<SnpPlatformStatus, UserApiError> {
-        let mut platform_status: SnpPlatformStatus = SnpPlatformStatus::default();
+        // Create an empty buffer for the SNP Platform Status to be written to by the Kernel.
+        let mut platform_status: FFISnpPlatformStatus = FFISnpPlatformStatus::default();
+
         let mut cmd_buf = Command::from_mut(&mut platform_status);
 
         SNP_PLATFORM_STATUS
             .ioctl(&mut self.0, &mut cmd_buf)
             .map_err(|_| cmd_buf.encapsulate())?;
 
-        Ok(platform_status)
+        // Determine SEV-SNP CPU generation in order to parse platform status accordingly.
+        let raw_cpuid = unsafe { std::arch::x86_64::__cpuid(0x8000_0001) }
+            .eax
+            .to_le_bytes();
+
+        let generation: Generation = raw_cpuid.as_slice().try_into()?;
+
+        Ok((generation, &*platform_status).try_into()?)
     }
 
     /// The firmware will perform the following actions:  
--- a/src/firmware/linux/host/ioctl.rs
+++ b/src/firmware/linux/host/ioctl.rs
@@ -9,9 +9,6 @@ use crate::impl_const_id;
 
 use crate::error::FirmwareError;
 
-#[cfg(feature = "snp")]
-use crate::firmware::host::SnpPlatformStatus;
-
 use std::marker::PhantomData;
 
 use iocuddle::*;
--- a/src/firmware/linux/host/types/snp.rs
+++ b/src/firmware/linux/host/types/snp.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use std::ops::{Deref, DerefMut};
+
 #[cfg(target_os = "linux")]
 use crate::error::CertError;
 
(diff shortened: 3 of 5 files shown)


src/firmware/host/mod.rs
```diff
@@ -38,6 +38,10 @@ use std::convert::TryInto;
 #[cfg(target_os = "linux")]
 use super::linux::host::types::SnpCommit;
 
+#[cfg(feature = "snp")]
+#[cfg(target_os = "linux")]
```
tylerfanelli, line 42:
```suggestion
#[cfg(all(target_os = "linux", feature = "snp"))]
```

tylerfanelli (changes requested):

src/firmware/host/mod.rs
```diff
@@ -38,6 +38,10 @@ use std::convert::TryInto;
 #[cfg(target_os = "linux")]
 use super::linux::host::types::SnpCommit;
 
+#[cfg(feature = "snp")]
```
tylerfanelli, line 41: This can be removed.

src/firmware/linux/host/types/snp.rs
```diff
@@ -314,6 +316,41 @@ impl<'a> std::convert::From<&WrappedVlekHashstick<'a>> for SnpVlekLoad {
     }
 }
 
+#[cfg(feature = "snp")]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+#[repr(C, packed)]
+/// Kernel friendly Snp Platform Status
```
tylerfanelli, line 322:
```suggestion
/// Kernel-friendly SNP platform status.
```

src/firmware/linux/host/types/snp.rs
```diff
@@ -314,6 +316,41 @@ impl<'a> std::convert::From<&WrappedVlekHashstick<'a>> for SnpVlekLoad {
...
+pub struct SnpPlatformStatus {
+    /// Entire Platform Status
```
tylerfanelli, line 324: This comment feels redundant.

src/firmware/linux/host/ioctl.rs
```diff
@@ -10,7 +10,7 @@ use crate::impl_const_id;
 use crate::error::FirmwareError;
 
 #[cfg(feature = "snp")]
-use crate::firmware::host::SnpPlatformStatus;
+use SnpPlatformStatus;
```
tylerfanelli, line 13:
```suggestion
use super::types::SnpPlatformStatus;
```

I think that would be the path? Try referring to the full path of `SnpPlatformStatus`.

tylerfanelli (changes requested):

src/firmware/linux/host/ioctl.rs
DGonzalezVillal: Yeah, isn't this imported in line 6. Can probably remove this

DGonzalezVillal (changes requested):
Only the one thing Tyler pointed out.

@larrydewey @tylerfanelli  The Platform Status struct is outdated (missing fields). I don't know if you guys want to address it in this PR or fix it in a separate one.

src/lib.rs
```diff
@@ -198,6 +198,53 @@ pub enum Generation {
     Turin,
 }
 
+#[cfg(all(not(test), feature = "snp"))]
```
tylerfanelli, line 201: What is the reason for `not(test)` being specified?

ajcaldelas: I'm not following isn't `not(test)` being specified with the additonal requirement that the snp is enabled?

I added this because with having only `not(test)` caused pipeline errors when the no-feature flag was enabled

larrydewey (dismissed):

src/firmware/host/mod.rs
```diff
@@ -209,14 +215,31 @@ impl Firmware {
     /// ```
     #[cfg(feature = "snp")]
     pub fn snp_platform_status(&mut self) -> Result<SnpPlatformStatus, UserApiError> {
-        let mut platform_status: SnpPlatformStatus = SnpPlatformStatus::default();
+        // Create an empty buffer for the SNP Platform Status to be written to by the Kernel.
+        let mut platform_status: FFISnpPlatformStatus = FFISnpPlatformStatus::default();
+
+        // Create a kernel-expected command structure which identifies what command to run, and what
```
tylerfanelli, line 221: I don't think this comment is necessary.

src/firmware/host/mod.rs
```diff
@@ -209,14 +215,31 @@ impl Firmware {
...
+        // buffer to write the result to.
         let mut cmd_buf = Command::from_mut(&mut platform_status);
 
+        // Communicate with the kernel, telling it to run the SNP_PLATFORM_STATUS command; updating the
```
tylerfanelli, line 225: Unnecessary comment.

src/firmware/host/mod.rs
```diff
@@ -209,14 +215,31 @@ impl Firmware {
...
 
-        Ok(platform_status)
+        // Independently find the generation of the CPU, as the
+        // SnpPlatformStatus has no context regarding the CPU generation, but
+        // the generation is needed to properly parse the SnpPlatformStatus.
+        let raw_cpuid = unsafe { std::arch::x86_64::__cpuid(0x8000_0001) }
+            .eax
+            .to_le_bytes();
+
+        // Try to turn the raw CPUID byte buffer into a Generation defined.
```
tylerfanelli, line 238: Unnecessary comment.

src/firmware/host/mod.rs
```diff
@@ -209,14 +215,31 @@ impl Firmware {
...
+        // Communicate with the kernel, telling it to run the SNP_PLATFORM_STATUS command; updating the
+        // command buffer with the result set.
         SNP_PLATFORM_STATUS
             .ioctl(&mut self.0, &mut cmd_buf)
             .map_err(|_| cmd_buf.encapsulate())?;
 
-        Ok(platform_status)
+        // Independently find the generation of the CPU, as the
+        // SnpPlatformStatus has no context regarding the CPU generation, but
+        // the generation is needed to properly parse the SnpPlatformStatus.
```
tylerfanelli, line 233:
```suggestion
        // Determine SEV-SNP CPU generation in order to parse platform status accordingly.
```

tylerfanelli (changes requested):

tylerfanelli:
How flexible is this? As I understand, Turin can use older attestation report versions. Is the same true with platform status?

src/lib.rs
tylerfanelli: As I understand, `not(test)` indicates that this won't be included if `cargo test` is being run. I think this can just be:

`#[cfg(feature = "snp")]`

tylerfanelli:
> @larrydewey @tylerfanelli The Platform Status struct is outdated (missing fields). I don't know if you guys want to address it in this PR or fix it in a separate one.

A separate one is fine with me.

larrydewey:
> How flexible is this? As I understand, Turin can use older attestation report versions. Is the same true with platform status?

Any underlying structure which has a `TCB_VERSION` will need to understand how to format properly for the various versions.

DGonzalezVillal:
> > @larrydewey @tylerfanelli The Platform Status struct is outdated (missing fields). I don't know if you guys want to address it in this PR or fix it in a separate one.
> 
> A separate one is fine with me.

We ended up throwing it in this one. This PR essentially fixed the TCB bug and updated the SnpPlatformStatus to have the missing data fields.

DGonzalezVillal:
@ajcaldelas If you could modify the commit message to be a little more descriptive of all the work we did, that would be appreciated.

tylerfanelli (approved):
LGTM, thanks!

PR #309, opened 2025-05-27
AttestationReport: Improving parsing and writing

larrydewey:
- Added two inline functions to reduce the duplication when parsing or writing attestation bytes.

- Only returing on instance of the object when parsing the bytes now.


DGonzalezVillal (approved):
LGTM

tylerfanelli (approved):
LGTM, thanks!

PR #312, opened 2025-06-19
Updating Library to Match Version 1.58 of the SNP Spec

DGonzalezVillal:
The new 1.58 firmware specification has been released, introducing changes to both the attestation report and key derivation features.

Attestation Report Changes:
Two new fields have been added: launch_mit_vector and current_mit_vector. These vectors reflect mitigations that may or may not be present on the system. Additionally, a new enablement bit for SEV-TIO has been added to the PlatformInfo structure. The attestation report version is now bumped to version 5. (Note: Version 4 remains under AMD embargo, so specifics cannot be disclosed.)

Key Derivation Changes:
A new field, launch_mit_vector, has been added to the key derivation request structure. This allows a mitigation vector to be mixed into the derived key. As a result, the message key request has been bumped to version 2. Version 2 requests require a mit_vector value; if a version 1 request is used, the mit_vector defaults to zero.

Parsing Improvements:
The attestation report byte parsing logic has been updated. Previously, there were separate variants for each version (pre-Turin and Turin). Now, the write_tcb and parse_tcb functions are used more effectively. Once the processor generation is determined, the appropriate TCB layout can be applied. Since TCB layout is the main difference between pre-Turin and Turin+ attestation reports, this change allows us to maintain a single report variant per version.

### DIFF
--- a/src/firmware/guest/types/snp.rs
+++ b/src/firmware/guest/types/snp.rs
@@ -11,6 +11,7 @@ use crate::{
         array::Array,
         parser::{ByteParser, ReadExt, WriteExt},
     },
+    Generation,
 };
 
 use serde::{Deserialize, Serialize};
@@ -18,10 +19,7 @@ use serde::{Deserialize, Serialize};
 use std::{fmt::Display, io::Write, ops::Range};
 
 #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
-use std::{
-    convert::TryFrom,
-    io::{self, ErrorKind},
-};
+use std::{convert::TryFrom, io};
 
 #[cfg(feature = "openssl")]
 use std::io::Error;
@@ -33,6 +31,8 @@ use openssl::{ecdsa::EcdsaSig, sha::Sha384};
 
 const ATT_REP_FW_LEN: usize = 1184;
 const CHIP_ID_RANGE: Range<usize> = 0x1A0..0x1E0;
+const CPUID_FAMILY_ID_BYTES: usize = 0x188;
+const CPUID_MODEL_ID_BYTES: usize = 0x189;
 
 /// Structure of required data for fetching the derived key.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
@@ -184,10 +184,8 @@ pub enum ReportVariant {
     V2,
 
     /// Version 3 of the Attestation Report for PreTurin CPUs.
-    V3PreTurin,
+    V3,
 
-    /// Version 3 of the Attestation Report for Turin+ CPUs.
-    V3Turin,
 }
 
 impl ReportVariant {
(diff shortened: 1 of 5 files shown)


DGonzalezVillal:
Updated SPEC: chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf

DGonzalezVillal:
Will fix failures now

src/firmware/guest/types/snp.rs
```diff
@@ -422,16 +454,25 @@ impl AttestationReport {
         // Determine the variant based on version and CPUID step
         let variant = if self.version == 2 {
             ReportVariant::V2
```
larrydewey, line 456: We could make this conditional block a bit more concise with a match statement.

```rust
let variant = match self.version {
    2 => ReportVariant::V2,
    3 | 4 => ReportVariant::V3,
    _ => ReportVariant::V5
};
```

larrydewey (approved):
One nit which isn't a requirement to be changed. Just a thought.

tylerfanelli (changes requested):
Can you break this change up into multiple commits? Each commit should include one logical change rather than one sweeping commit listing multiple changes in the description.

DGonzalezVillal:
@tylerfanelli Please take a look now

tylerfanelli (approved):

cschramm:
In the report that I provided in hex, you can see that the version byte (offset 0) indicates 3, and family (offset 0x188) and model (offset 0x189) are both 0. `AttestationReport::from_bytes` passes those values to `Generation::identify_cpu` which fails with the `processor is not of know SEV-SNP generation.` error.

I have no idea how that report came to be as it does not match the specification, but we have multiple such reports in our fixtures. They are valid and should have come from an EPYC 9454P in January. The machine received a firmware update in the meantime and we are not able to replicate such reports.

DGonzalezVillal:
There is a chance those reports were either parsed incorrectly by old tooling or some firmware error? 

Version 3 of the report should have a valid family ,model and stepping.

cschramm:
We can rule out incorrect parsing as a source, as the reports have a signature that verifies fine with their corresponding VCEK certificate (https://kdsintf.amd.com/vcek/v1/Genoa/9dc99962c063029e430b6f7b734075ec542f4f3ec639e657e14585d3fe559b38532bc42d037d618317694ad6634d2507964e276c8eedeac4978c7006bf89f8e1?blSPL=9&teeSPL=0&snpSPL=22&ucodeSPL=72), e.g. using `sev` 6.2's parser and verification. As the signature covers both the version 3 byte and the 0 family, model, and stepping bytes, the report must have been issued exactly like that.

Anyway, if this weird phenomenon is something completely new to you, it's probably not worth handling. We do not have a use case either, just noticed it in our fixtures, which we can simply replace to make them work with sev 6.3 and 7.

PR #316, opened 2025-07-22
Refactor/Fix: Restoring snp_vlek_load api to prior release and fixing functionality

DGonzalezVillal:
Restoring snp_vlek_load function to avoid breaking api. It takes the usual slice with the hashstick instead of the new strucutre. The new structure remains, since it can be used to validate the hashstick bytes.

### DIFF
--- a/src/firmware/linux/host/types/snp.rs
+++ b/src/firmware/linux/host/types/snp.rs
@@ -251,23 +251,23 @@ impl Default for SnpSetConfig {
     }
 }
 
-// Length defined in the Linux Kernel for the IOCTL.
+// Expected length for the VLEK hashstick.
 const HASHSTICK_BUFFER_LEN: usize = 432;
 
 #[cfg(feature = "snp")]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[repr(C, packed)]
-/// Wrapped VLEK data.
-pub struct WrappedVlekHashstick<'a> {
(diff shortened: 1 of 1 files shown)


DGonzalezVillal:
I was finally able to get my hands on a VLEK and a VLEK hashstick, so I started testing VLEK loading and realized that our current implementation was broken. I’ve updated this PR with changes that reflect what I found to work.

First, regarding the FFI WrappedVlekStructure: we were passing a reference to this struct in the command buffer, but the kernel expects the struct to own the data. I tried to adjust the existing implementation and play around with lifetimes, but ultimately found that making the struct own the data—rather than referencing it—was the simplest and most reliable solution.

Next, in the SnpVlekLoad structure, the length parameter was incorrectly set to the length of the hashstick. In reality, this value should be set to the length of the structure itself (16 bytes).

With these changes, I was able to successfully load a hashstick on a host. The original API can stay the same, so it's not changed for users.

Also worth noting: the upstream kernel currently has a bug I identified, which prevents the feature from working correctly even though it is technically enabled. It will require a patch for proper functionality.

Let me know your comments or suggestions.

@tylerfanelli @larrydewey

tylerfanelli (approved):

tylerfanelli:
LGTM. I feel that the VLEK hashstick feature is so new that some instability is expected, correct?

larrydewey (approved):
This looks good, but I do think we should adhere to semver and change the version before releasing.

PR #318, opened 2025-07-29
Bumping crate version

DGonzalezVillal:
@larrydewey @tylerfanelli 

After https://github.com/virtee/sev/pull/316 get merged, we want to cut a new release. The question is if we want to cut 6.3.0 or 7.0.0? 
PR https://github.com/virtee/sev/pull/316 is going to introduce some API changes, but as far as I can tell, that API has never been functional—some of the structure initializations needed changes to actually make it work. I suspect it hasn't been widely used because of this, although we have had people reach out asking how to perform the VLEK load.
The question is: do we need to bump the version to 7.0 after this change, or would it be acceptable to move forward with a 6.3 release instead? We’d prefer to avoid another major release at this time, so we’re checking in to hear everyone’s thoughts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1606,7 +1606,7 @@ dependencies = [
 
 [[package]]
 name = "sev"
-version = "6.2.1"
+version = "6.3.0"
 dependencies = [
  "base64",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sev"
-version = "6.2.1"
+version = "6.3.0"
 authors = [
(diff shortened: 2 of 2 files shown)


mkulke:
It's a good policy to run cargo semver-check to understand whether a release introduced a breaking change. In general if exported types/signatures have changed, I would consider a major bump, otherwise things might start to be complex for crate consumers, which have both direct and indirect dependencies on the sev crate.

tylerfanelli:
I was initially against it, but if cargo semver-check indicates a change, then it's best to update. Plus we now have #319 which will be a breaking change anyway, so we can just update to 7.0.0 when that is merged.

PR #320, opened 2025-08-06
Refactor: Chaning vlek_load api to thake in WrappedVlekHashstick

DGonzalezVillal:
Since we're updating the major version, I'd prefer having users pass the UAPI structure to the API instead of a slice. By using the UAPI struct, the library can perform the appropriate length and reserved field checks before passing the data to the firmware. This gives us better validation of the data upfront.

This approach is cleaner than passing a raw slice and makes it easier to work with compared to manually managing arrays or buffers.


larrydewey (approved):
LGTM

Issue #323, opened 2025-08-12
External Crates Exposed in our library


DGonzalezVillal:
As @mkulke pointed out, we have types external to our crate that are exposed publicly. Using [cargo-check-external-types](https://github.com/awslabs/cargo-check-external-types), I generated a list of these types. I’m sharing it here to start a discussion on how to handle them.

For example, error types could be replaced with strings, but things like Serialize and Deserialize aren’t as straightforward — we may need to consider keeping those exposed.

Here’s the list:

| Crate | Type | Used In |
| ---   | ---  | ---     |
| bincode | bincode::error::ErrorKind | src/error.rs:1003:18 |
| bincode | bincode::error::ErrorKind | src/error.rs:1098:1 |
| bincode | bincode::error::ErrorKind | src/error.rs:725:18 |
| bincode | bincode::error::ErrorKind | src/error.rs:754:1 |
| bincode | bincode::error::ErrorKind | src/error.rs:929:18 |
| bincode | bincode::error::ErrorKind | src/error.rs:981:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:765:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:765:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:894:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:894:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:95:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:95:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:985:1 |
| bitfield | bitfield::BitRange | src/firmware/guest/types/snp.rs:985:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:228:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:228:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:31:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:31:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:558:1 |
| bitfield | bitfield::BitRange | src/firmware/host/types/snp.rs:558:1 |
[...]

PR #324, opened 2025-08-13
Refactor: removing CODICON crate dependency

DGonzalezVillal:
Codicon has not been maintained for 5 years and only defines two traits for encoding and decoding types. We were exposing these traits externally, forcing users of virtee to also depend on Codicon. By inlining the traits into our main library, we can keep exposing them publicly while removing the Codicon dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,12 +146,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
-name = "codicon"
-version = "3.0.0"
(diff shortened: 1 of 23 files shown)


tylerfanelli:
Seems there's clippy + fmt errors. Can you address them?

DGonzalezVillal:
@tylerfanelli so from this list right here:
https://github.com/virtee/sev/issues/323

We are ok with exposing the traits for Bincode, serde, openssl and uuid?

Or how do we want to approach this. I was working on other PRs to stop exposing traits from these other crates, but it seems to me that it won't really be possible to not expose ANY of these traits. 

Maybe make a comprehensive list of what we are ok and we aren't exposing?

tylerfanelli (approved):

PR #328, opened 2025-08-25
New Parsing library and remove Bincode and Codicon dependencies

DGonzalezVillal:
This PR introduces an internal way to parse bytes for our structures.

It introduces the new traits Encode and Decode, which are used to to describe encoding and decoding relationships for our structures. 

We then use this traits to extend std Read and Write in order to be able to read/write the bytes exactly how we need.

ByteParser then becomes a trait that can be used to write and read bytes.

This addition helps us completely remove our dependency on Bincode.


src/certs/sev/chain.rs
```diff
@@ -5,6 +5,7 @@
 use super::*;
 
 use serde::{Deserialize, Serialize};
+// use std::result::Result;
```
tylerfanelli, line 8: Remove.

tylerfanelli (approved):
LGTM. one comment and we can merge and cut a 7.0.0 release

larrydewey (approved):

PR #329, opened 2025-08-26
snp: skip calling CPUID on non-x86 platform

### LINKED ISSUE #326: Crate cannot be built on non-x86 platforms
The crate uses the `std::arch::x86` code [here](https://github.com/virtee/sev/blob/8914fd3213eccb3f12e5a1aa30bd0b009801c9d7/src/lib.rs#L289) which will not build on other platforms.

In Trustee we use the SEV crate to help us evaluate SEV evidence on any platform.

This cpuid call should probably be behind a flag so that it is only compiled when on x86.

fitzthum:
When this crate is used to parse SNP hardware evidence on a non-x86 platform, it will still try to build with cpuid from the x86 crate, which will not be found.

Instead, put cpuid behind a platform config flag. If we are not on x86, don't call cpuid and return an error.

Fixes #326 

As an aside, it seems kind of odd that we call this from the types code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -286,10 +286,17 @@ impl Generation {
     pub fn identify_host_generation() -> Result<Self, std::io::Error> {
         use std::convert::TryInto;
 
-        let raw_cpuid = unsafe { std::arch::x86_64::__cpuid(0x8000_0001) }
+        #[cfg(target_arch = "x86_64")]
+        return unsafe { std::arch::x86_64::__cpuid(0x8000_0001) }
             .eax
-            .to_le_bytes();
-        raw_cpuid.as_slice().try_into()
+            .to_le_bytes()
+            .as_slice()
+            .try_into();
+
+        #[cfg(not(target_arch = "x86_64"))]
+        Err(std::io::Error::other(
+            "Cannot get EPYC generation on non-x86 platform",
+        ))
     }
 }
 


tylerfanelli:
It's called from the types code because we don't expect a user to know which exact SNP generation (Milan, Genoa, Turin, any future ones) their running on. As such, we determine it for them when they go to set data that is dependent on the underlying generation.

tylerfanelli (approved):

fitzthum:
I guess it just seems like we're mixing types, which could be used by basically anyone, and some of the platform-specific logic. I haven't looked through this in a while though and I have no suggestions.

DGonzalezVillal:
@fitzthum Did you test and see if this fixed the issue? 

This functionality is used inside some HOST SNP APIs, so I want to make sure that this is the only thing needed to get the compilation to work on ARM. APIs: https://github.com/virtee/sev/blob/main/src/firmware/host/mod.rs

I was thinking of creating a dummy function that replaces this cpuid stuff when the library is compiled a non-x86_64 machine.

Something like:

```rust
   #[cfg(any(not(feature = "snp"), not(target_arch = "x86_64")))]
    pub fn identify_host_generation() -> Result<Self, std::io::Error> {
        Err(std::io::Error::new(
            std::io::ErrorKind::Unsupported,
            "SNP host generation detection only supported on x86_64 targets",
        ))
    }
```

fitzthum:
I don't have machine to test it on, but we can use https://github.com/confidential-containers/trustee/pull/931 to see if it passes the CI. This is basically blocking all our PRs right now so would be good to fix.

fitzthum:
Trustee CI can build on ARM now https://github.com/confidential-containers/trustee/actions/runs/17279027197/job/49043047711?pr=931

We do have another potential ARM issue that may or may not be related. Test hasn't finished for that yet.

tylerfanelli:
Just to be clear, you mean the Trustee CI can build **with this change in place**, correct? If so, @DGonzalezVillal let's merge this and push a minor release (6.3.1) to crates to unblock them.

tylerfanelli (approved):

PR #342, opened 2025-09-15
Fix: Adding serde_bytes as serde dependency

DGonzalezVillal:
Serde_bytes was accidentally removed, it is still needed as one of the serde dependencies.


tylerfanelli:
Closed as this was addressed in #342

PR #356, opened 2025-11-13
Add From implementation to Certificate when cert_nossl

BRA1L0R:
The `Certificate` implementation under cert_nossl.rs (the one used when the cert_nossl feature is enabled) is missing a From implementation between the `Certificate` wrapper and the inner `x509_cert::Certificate`

I also bumped the minor version as this does not break current APIs.

On a side note, the currently implemented From<&Certificate> (and the other way around) in the openssl implementation are wrong. Cloning should not be done implicitly inside From implementation.

I addede a cert method returning a reference to the inner type to support cloning explicitly the x509_cert certificate type instead of cloning it implicitly via `.into()`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 .vscode/
+.zed/
--- a/src/certs/snp/cert_nossl.rs
+++ b/src/certs/snp/cert_nossl.rs
@@ -62,6 +62,11 @@ impl Verifiable for (&Certificate, &Certificate) {
 }
 
 impl Certificate {
+    /// Gets a reference to the X509 certificate inside
+    pub fn cert(&self) -> &x509_cert::Certificate {
+        &self.0
+    }
+
     /// Create a Certificate from a PEM-encoded X509 structure.
     pub fn from_pem(pem: &[u8]) -> Result<Self> {
         let cert = x509_cert::Certificate::from_pem(pem)
@@ -105,4 +110,16 @@ impl Certificate {
 
 fn io_error_other<S: Into<String>>(error: S) -> io::Error {
     io::Error::new(ErrorKind::Other, error.into())
+}
+
+impl From<x509_cert::Certificate> for Certificate {
+    fn from(value: x509_cert::Certificate) -> Self {
+        Self(value)
+    }
+}
+
+impl From<Certificate> for x509_cert::Certificate {
+    fn from(Certificate(cert): Certificate) -> Self {
+        cert
+    }
 }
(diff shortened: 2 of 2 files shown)


DGonzalezVillal (changes requested):

DGonzalezVillal:
The code itself looks good although. Please make sure you add your sign to the commits and also please  raise an issue regarding the &Certificate From implementation.

DGonzalezVillal:
@BRA1L0R were you able to put the requested changes in?

DGonzalezVillal (approved):

src/certs/snp/cert_nossl.rs
```diff
@@ -62,6 +62,11 @@ impl Verifiable for (&Certificate, &Certificate) {
 }
 
 impl Certificate {
+    /// Gets a reference to the X509 certificate inside
+    pub fn cert(&self) -> &x509_cert::Certificate {
+        &self.0
+    }
+
```
DGonzalezVillal: We only have a problem with this function. The From implementations are fine. 

This function is returning a ref to an external type, and we want to avoid that. Also we don't have an equivalent function to this on the openssl implementation.

In openssl we do have a function returning the Pkey type, so that should be addressed, but no_ssl already has this implementation. So ideally we remove this and keep the from.

DGonzalezVillal (changes requested):

src/certs/snp/cert_nossl.rs
BRA1L0R: Totally fine, sorry for the misunderstanding. The `From` implementations are fine for my usecase, however the current API does not allow accessing the certificate without destructuring the Certificate object. That method was supposed to give borrowed access to it.

DGonzalezVillal (approved):

tylerfanelli (commented):
We are exposing types from other crates. This introduces a dependency issue. Not sure if we want to do that.

DGonzalezVillal (approved):

tylerfanelli (approved):

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
-.vscode/
\ No newline at end of file
+.vscode/
+.zed/
```
tylerfanelli, line 3: What is this?

tylerfanelli:
Please add a `Signed-off-by` line to the commits.

DGonzalezVillal:
@BRA1L0R sorry for the long delay, if you could solve the top two concerns we can merge as is

.gitignore
larrydewey: Zed is an editor like VSCode. This is to make sure those settings are synced into the repository

larrydewey (approved):

PR #362, opened 2025-12-09
chore: allow validating SNP measurement hashes in non-linux OS

mfontanini:
This allows all symbols that are needed for SNP measurrement hash generation (e.g. everything needed to invoke `snp_calc_launch_digest`) to be used in non linux OSes. The main thing preventing this seemed to be `launch::snp::PageType` since all of `launch::snp` was gated behind `target_os = "linux"`. I moved this type up to `launch`, which is maybe not ideal. I'm open to alternatives such as creating 2 modules within `launch::snp` so one contains this `PageType` (and maybe something else?), or anything else that makes sense.

### DIFF
--- a/src/launch/snp.rs
+++ b/src/launch/snp.rs
@@ -4,7 +4,6 @@
 //! This ensures (at compile time) that the right steps are called in the
 //! right order.
 
-use crate::launch::PageType;
 #[cfg(target_os = "linux")]
 use crate::{
     error::FirmwareError,
@@ -18,6 +17,8 @@ use bitflags::bitflags;
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
+
+pub use crate::launch::PageType;
 
 /// Launcher type-state that indicates a brand new launch.
 pub struct New;


DGonzalezVillal:
Hello @mfontanini thank you for the PR. I don't love the idea of moving the PageType to launch. I could give better suggestions if I understood your use case a little better. 

What I'm imagining is that you are trying to generate hashes in a non-linux environment but, you can't because the measurement library is not compiling because of the linux flag. Is that correct?

DGonzalezVillal:
I see. I'm thinking you've hit in a bigger issue, where maybe we are using this gating functions in the wrong way. 
I think maybe only the IOCTL API's should be gated behind the linux flag, since you won't be able to reach those in a non-linux machine. But everything that is parsing, validating or calculating should be usable on any platform.

For this particular PR, you can remove the Linux gate on the measurement library and that should unblock your use case. That way you don't need to move PageType. I'll add to the to-do list to make sure other functionality isn't being blocked this way.

DGonzalezVillal (approved):
Wait I understand what this is doing better now. This is a good solution. Thoughts @tylerfanelli

src/launch/snp.rs
```diff
@@ -4,6 +4,7 @@
 //! This ensures (at compile time) that the right steps are called in the
 //! right order.
 
+use crate::launch::PageType;
```
mfontanini: FYI if this solution works this should probably be a `pub use` so this doesn't break backwards compatibility.

DGonzalezVillal: I don't think that would be the case because PageType is now defined in mod right? we are not exporting it from launch::snp, but from launch directly, if my reasoning is correct.

DGonzalezVillal:
But now that I think about it. Would it make more sense to move PageType to measurement and have launch import it. Because PageType will get compiled regardless now

DGonzalezVillal:
@mfontanini  Were you able to make the changes avoiding the breaking change?

DGonzalezVillal (approved):
Thank you! @tylerfanelli PTAL

PR #363, opened 2025-12-17
Lax parser feature

### LINKED ISSUE #343: Forward compatibility
In a remote attestation use case, a client needs to parse attestation reports provided by a server. The `sev` crate currently does not seem like a good fit for that, as it is not forward compatible.

For example, an existing client fails to parse a V5 report provided by a server if it has not received #312 yet. This seems rather unnecessary and the client could easily provide forward compatibility as long as new attestation versions do not actually break previous specs.

The reason that the parser fails is simply because `skip_bytes` expects all reserved areas to be zeroed. The spec does not even require that for all reserved areas. It seems like the only reason for the parser to expect it is that when a parsed `AttestationReport` gets encoded to bytes, reserved areas also simply get filled with zeros.

Would you consider actually reading and writing contents of reserved areas instead of requiring them to be zeroed to provide simple forward compatibility for distributed software?

cschramm:
The parser strictly checks reserved areas to be zero (even though the specification does not even require that for all of them). This is an issue when parsing future (i.e., not known to the current codebase) versions of attestation reports, where those areas may no longer be reserved.

The `lax-parser` feature disables that check, currently with some tradeoffs:

* No `WriteExt::skip_bytes`
* No `Encoder` `impl` for `AttestationReport`, `Signature`, and `WrappedVlekHashstick`
* No compatibility with `openssl` and `crypto_nossl`

Closes #343

### DIFF
--- a/src/firmware/host/mod.rs
+++ b/src/firmware/host/mod.rs
@@ -286,6 +286,7 @@ impl Firmware {
         Ok(())
     }
 
+    #[cfg(not(feature = "lax-parser"))]
     #[cfg(feature = "snp")]
     /// Insert a Version Loaded Endorsement Key Hashstick into the AMD Secure Processor.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,6 +90,14 @@ compile_error!(
     "feature \"openssl\" and feature \"crypto_nossl\" cannot be enabled at the same time"
 );
 
+#[cfg(all(
+    any(feature = "openssl", feature = "crypto_nossl"),
+    feature = "lax-parser"
+))]
+compile_error!(
+    r#"feature "openssl" and feature "crypto_nossl" cannot be used together with "lax-parser""#
+);
+
 /// SEV and SEV-SNP certificates interface.
 pub mod certs;
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -213,3 +213,28 @@ jobs:
             flag: --release
         features:
           - crypto_nossl
+
+  sw-lax-parser:
+    name: sw lax-parser ${{ matrix.runner }} ${{ matrix.toolchain }} ${{ matrix.profile.name }} ${{ matrix.features }}
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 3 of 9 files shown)


DGonzalezVillal:
Because the Write stuff can be modified to write the raw contents. Create something like a raw write

src/firmware/host/mod.rs
```diff
@@ -286,6 +286,7 @@ impl Firmware {
         Ok(())
     }
 
+    #[cfg(not(feature = "lax-parser"))]
```
DGonzalezVillal: Why do you have to disable this API for lax-parser?

src/lib.rs
```diff
@@ -90,6 +90,14 @@ compile_error!(
     "feature \"openssl\" and feature \"crypto_nossl\" cannot be enabled at the same time"
 );
 
+#[cfg(all(
+    any(feature = "openssl", feature = "crypto_nossl"),
+    feature = "lax-parser"
+))]
```
DGonzalezVillal: Why would lax-parser be part of this check?

DGonzalezVillal (changes requested):
Hello @cschramm,

Apologies for the delay—I was out on break.

After taking a closer look at your PR, I’m thinking a better approach might be to shift from the current lax-parsing idea to something more like a raw read/write (or unsafe read/write) mode. With that approach, this feature would simply read and write the raw binary contents as-is.

In our earlier discussions, I may have implied that I wanted this scoped strictly to parsing, but after reviewing the PR and realizing that lax parsing would require disabling or weakening other APIs, I’m less comfortable with that direction. I’d prefer not to reduce existing functionality to support a single feature. This alternative also results in fewer changes overall, since it would mainly involve updating the read and write paths to operate on raw bytes without safety checks.

This keeps the library behavior unchanged for most users, while clearly shifting responsibility for validation and safety checks to users who opt into this mode.

Please let me know what you think, and apologies again for the back and forth.

Thanks!

cschramm:
I see. `AttestationReport`'s `Decoder` would then use those getters? How would it apply the zero checks? Raw access (applying structural knowledge)? A method on the raw type dedicated to those checks? :thinking:

Fine for me either. @DGonzalezVillal?

cschramm:
Closing in favor of #372 (and #378)

PR #366, opened 2026-01-07
Lax parser feature for attestation reports

### LINKED ISSUE #343: Forward compatibility

cschramm:
The parser strictly checks reserved areas to be zero (even though the specification does not even require that for all of them). This is an issue when parsing newer versions of attestation reports where those areas might not be reserved anymore. The `lax-parser` feature disables that check.

`AttestationReport` actually reads and writes reserved bytes now.

Closes #343

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -213,3 +213,28 @@ jobs:
             flag: --release
         features:
           - crypto_nossl
+
+  sw-lax-parser:
+    name: sw lax-parser ${{ matrix.runner }} ${{ matrix.toolchain }} ${{ matrix.profile.name }} ${{ matrix.features }}
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


cschramm:
Closing in favor of #372 (and #378)

PR #370, opened 2026-01-20
Unsafe-Parser 

DGonzalezVillal:
Introduce a new `unsafe_parser` feature that removes the `skip_bytes`
logic from extended read and write operations in the parser. This
forces structures to be read and written using the exact raw contents
present in the input, avoiding issues with unparsable reports when new
report versions are released.

This change removes behind-the-scenes version-specific parsing of
attestation reports and shifts responsibility to the user to validate
and interpret report contents during attestation and related workflows.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,6 +34,17 @@ jobs:
           toolchain: 1.85.0
       - run: cargo clippy --features=crypto_nossl,hw_tests,dangerous_hw_tests --all-targets -- -D clippy::all -D unused_imports -D warnings -D clippy::style
 
+  clippy-unsafe-parser:
+    name: cargo clippy unsafe-parser
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


DGonzalezVillal:
> At this point, I can't really meaningfully review this code - lack of familiarity, still learning Rust, etc
> 
> Just the one comment on use of literals, and not Rust constants, for field sizes.
> 
> I will say, in general, that interpretation of attestation report bytes on the attesting platform isn't necessary "in production". It's only necessary on the verifier and it should be easier to update verifiers than attesting platforms since far fewer verifiers. Given my current lack of familiarity, I don't know how this code relates to production attesting platform or verifier code.

You bring up a good point. Attesting platforms don't NEED to know what the report values are, but verifiers do need to be able to parse this report. So this goes back to what @hyperfinitism is saying, of handling the raw blobs, and only converting when absolutely needed.

Issue #369, opened 2026-01-16
default feature "sev" prohibits compilation on non x86_84 targets


mkulke:
Is this intentional? in trustee there are some target archs (aarch64, s390) that would be able to verify SNP evidence, but you cannot currently build the "sev" feature on those platforms, because "arch::_cpuid" is used, which is x86 specific. This means a consumer has to disable default-features for this crate. However, if a transitive dependency on the sev crate is introduced and that one doesn't disable default-features, the sum of all features will be enabled, and compilation will fail. E.g.

```
trustee.deps:
- sev v=7, default-features=false,features=["serde"]
- some-crate v=1

some-crate.deps:
- sev v=7, default-features=true,

=> trustee.deps.sev.features = [default.., serde]
``` 

if it's unintentional, maybe a ci job w/ default features on an arm runner could help?

cclaudio:
Some time ago @fitzthum pushed the https://github.com/virtee/sev/pull/329 to fix a cpuid call. It looks like that the `rdrand` dependency also should to be behind a `x86_64` arch flag and fallback to `openssl` when in non-x86 arch.

With that, trustee would be able to run the SNP verifier in a non-x86 platform and use the sev crate to verify the SNP attestation report.

mythi:
> It looks like that the `rdrand` dependency also should to be behind a `x86_64` arch flag and fallback to `openssl` when in non-x86 arch.

That was covered by #340. AFAUI the remaining question is if `sev` default feature is intentional or not (this issue).

mkulke:
> That was covered by [#340](https://github.com/virtee/sev/pull/340). AFAUI the remaining question is if `sev` default feature is intentional or not (this issue).

exactly. do we want this work on non-x86_64 systems...

```bash
cargo init
cargo add sev
cargo build
```

or are the consumers of the crate required to find a permutation of features that will work on their target?

If we want the former, this could be ensured by putting those 3 lines in a workflow w/ `runs-on: ubuntu-24.04-arm`

PR #372, opened 2026-01-29
feat: introduce unsafe_parser

DGonzalezVillal:
When this feature is enabled, the zero-byte check performed by skip_bytes is disabled. As a result, parsing will no longer fail when fields that are expected to be zero contain non-zero values.

This enables forward compatibility when reserved fields are repurposed in newer report versions. Users enabling this feature must take care to validate reserved fields manually if they rely on their contents for security or correctness.

The approaches of preserving values in reserved fields or treating the report as an opaque blob both have potential. However, based on concerns raised by @cschramm and @hyperfinitism, I think there is a larger design discussion to be had—one that could ultimately change how some of these structs are built and may require a breaking API release. To avoid dependency issues like the one @cshramm pointed out in https://github.com/virtee/sev/pull/370
, I’m hesitant to put such changes behind a feature flag. The alternative—introducing a separate report type—is also not ideal.

I don’t want to block @cshramm any longer, so this change immediately addresses the forward compatibility problem. I’m aware this approach has trade-offs, such as losing access to the parsed bytes of invalid reserved fields. However, since we currently have no appropriate fields to store that data, the report maintains type safety by only exposing known, validated fields. Additionally, the API already returns the raw binary report, so those bytes can still be inspected if needed.

If there are additional functions or features you think would make this more immediately usable, please let me know. I’ll start a separate discussion to explore how we might incorporate @hyperfinitism’s ideas into a future major release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ sev = ["dep:rdrand"]
 snp = []
 crypto_nossl = ["dep:p384", "dep:rsa", "dep:sha2", "dep:x509-cert"]
 serde = ["dep:serde", "dep:serde-big-array", "dep:serde_bytes"]
+unsafe_parser = []
 
 [target.'cfg(target_os = "linux")'.dependencies]
 iocuddle = "^0.1"
--- a/src/util/parser_helper/read_ext.rs
+++ b/src/util/parser_helper/read_ext.rs
@@ -33,6 +33,7 @@ pub trait ReadExt: Read {
             while remaining > 0 {
                 let n = remaining.min(CHUNK);
                 self.read_exact(&mut buf[..n])?;
+                #[cfg(not(feature = "unsafe_parser"))]
                 if buf[..n].iter().any(|&b| b != 0) {
                     return Err(std::io::Error::new(
                         std::io::ErrorKind::InvalidData,


cschramm (approved):
Thank you, @DGonzalezVillal. Your push to release any solution to unblock me is very kind.

This very minimal solution is perfectly fine for me.

Just bare in mind that the `Encoder` impls still assume that reserved bytes are zero, leading to invalid binary and broken features that rely on it, especially the cryptographic verification. #363 thus disables those parts while #366 reads and writes reserved areas to avoid that (leading to overhead even without the feature enabled).

I'd personally favor the #363 approach to avoid the footgun that comes with the feature otherwise, but that's up to you.

DGonzalezVillal:
@tylerfanelli Are you ok with merging this or PR https://github.com/virtee/sev/pull/363?

I think after that we cut a minor release, because the more complete solution in https://github.com/virtee/sev/pull/378 will have major changes.

For the next major release we can also address the concerns in https://github.com/virtee/sev/issues/369

thoughts??

tylerfanelli (approved):
LGTM.

tylerfanelli:
@DGonzalezVillal Is there anything delaying this?

DGonzalezVillal:
Hey @cschramm sorry for the delay, handling several things at the same time now. 

Do you prefer merging this or waiting for https://github.com/virtee/sev/pull/378

I think 378 is the more complete solution, but I don't know how pressing this is.

cschramm:
An official (read: merged and ideally even released) short-term solution would be great.

I have to admit that I still have to check if the latest state of #378 works out (and I do have some doubts, as it looks like the parser comes with "validation" for reserved areas again)

DGonzalezVillal:
@cschramm I would appreciate your comments on the other PR, since I believe it's close to being done, I might merge that and keep the original AttestationReport with a deprecation warning.

A thing about the other solution, is that regardless of the version of the report you can alwasy validate the signature, but you won't be always be able to parse the body if support has not been added in

Issue #330, opened 2025-08-29
Misleading display of new V5 attestation report fields (`TIO_EN`, `PAGE_SWAP_DISABLE`) on V3/V4 reports


hyperfinitism:
When parsing and displaying attestation reports, the `sev` crate currently shows newly introduced bitfields `TIO_EN` (in `PLATFORM_INFO`) and `PAGE_SWAP_DISABLE` (in `GUEST_POLICY`) (both added in ABI Rev 1.58) as `false` for older reports (e.g. V3, V4).

This behavior can be misleading because these fields did not exist in earlier versions of the attestation report. Displaying them as `false` may suggest that the feature is *explicitly disabled*, while in reality the field is simply not present.

These fields should be represented as `None` rather than `false` when parsing V3/V4 reports. This would be consistent with how other fields introduced in ABI Rev 1.58 (`LAUNCH_MIT_VECTOR` and `CURRENT_MIT_VECTOR`) are currently handled.

### Example
- Azure DCasv6 (Genoa)
- OS: Ubuntu 24.04 LTS (CVM)
- Kernel: 6.11.0-1018-azure

<details>

<summary>Attestation Report of V3 (selected)</summary>

```plaintext
Attestation Report:

Version:                      3

Guest SVN:                    65547

Guest Policy (0x3001f):
  ABI Major:     0
  ABI Minor:     31
  SMT Allowed:   true
  Migrate MA:    false
  Debug Allowed: false
  Single Socket: false
  CXL Allowed:   false
  AEX 256 XTS:   false
  RAPL Allowed:  false
  Ciphertext hiding: false
  Page Swap Disable: false

...

  Microcode:   84
  SNP:         23
  TEE:         0
  Boot Loader: 10
  FMC:         None

Platform Info (39):
  SMT Enabled:               true
  TSME Enabled:              true
  ECC Enabled:               true
  RAPL Disabled:             false
  Ciphertext Hiding Enabled: false
  Alias Check Complete:      true
  SEV-TIO Enabled:           false

...

Reported TCB:

TCB Version:
  Microcode:   84
  SNP:         23
  TEE:         0
  Boot Loader: 10
  FMC:         None

...

Launch Mitigation Vector:     None

Current Mitigation Vector:    None

...
```

</details>

DGonzalezVillal:
Hello @hyperfinitism I have attempted to resolve this issue in this PR:
https://github.com/virtee/sev/pull/378

[...]

PR #378, opened 2026-02-04
New report parsing 

DGonzalezVillal:
## Overview

This PR implements a new SEV-SNP attestation report parsing flow, based on the discussion in  
https://github.com/virtee/sev/discussions/373.

The goal is to make it difficult to accidentally treat untrusted bytes as a verified report body, while still supporting efficient, zero-copy parsing and forward-compatible handling of newer firmware layouts.

---

### 1) `Report`: a zero-copy view of an *unverified* attestation report

`Report` borrows the raw report buffer and exposes the relevant slices:

- `body`: the signed portion of the report
- `signature`: the report signature bytes

`Report` is intentionally **not trusted**. It is only a framing layer that allows:

- Preserving the raw firmware report exactly as received
- Passing reports between intermediaries without parsing
- Performing signature verification without interpreting untrusted fields
- Maintaining forward compatibility for future report versions

---

### 2) `ReportBody`: a zero-copy, typed view of the report body

`ReportBody` parses the body into strongly typed fields while borrowing fixed-size arrays directly from the original buffer.
`ReportBody::from_bytes()` performs **no cryptographic verification**. It only validates structural layout and reserved fields.

The correct and safe way to obtain a `ReportBody` is via:

```rust
let report = Report::from_bytes(&raw_bytes)?;
let body = ReportBody::try_from((&report, &certificate_or_chain))?;
```

### The `TryFrom` Implementation

The `TryFrom` implementation:

- Verifies the report signature using the provided VEK or certificate chain.
- Only after successful verification, parses the signed body into a typed `ReportBody`.

This ensures all parsed fields are cryptographically authenticated.

---

### 3) `SignatureAlgorithm`: Single Entrypoint for Report Verification

This PR introduces a `SignatureAlgorithm` enum and makes it the single dispatch point for attestation report signature verification.

Changes include:

- Parsing the signature algorithm from `0x34..0x38` into `SignatureAlgorithm`
- Storing it in `Report`
- Delegating verification in `Verifiable for (&Certificate, &Report)` to:
```rust
SignatureAlgorithm::verify(body, signature, vek)
```
- Removing backend-specific hashing and signature conversion logic from firmware parsing code

This cleanly separates:
- Report layout/parsing
- Cryptographic verification logic

### 4) Spec-Driven Reserved/MBZ Validation and Stricter Parsing
Reserved-field handling is tightened to match the SNP ABI:
`GuestPolicy`
- Enforces RMB1 bit 17
- Validates MBZ (must-be-zero) reserved ranges

`PlatformInfo`
- Validates reserved bits

`TCB`
- Parsing updated to align strictly with generation-aware layouts
- Invalid reserved fields now fail early instead of being silently accepted.

### Intended Flow

```Plain Text
raw bytes
   ↓
Report::from_bytes (framing only)
   ↓  (signature verification via Certificate or Chain)
ReportBody::try_from((&report, &vek_or_chain))
   ↓
typed, verified ReportBody
```

### Forward Compatibility and Safety
This design:
- Preserves full raw report bytes
- Avoids lossy parsing of unknown or reserved fields
- Prevents unsafe coercion of future report versions
- Makes cryptographic verification explicit and centralized
- Ensures spec violations (reserved bits, MBZ fields, etc.) fail loudly

### DIFF
--- a/src/certs/snp/signature/ecdsa.rs
+++ b/src/certs/snp/signature/ecdsa.rs
@@ -61,6 +61,8 @@ impl Decoder<()> for Signature {
     fn decode(reader: &mut impl Read, _: ()) -> Result<Self> {
         let r = reader.read_bytes()?;
         let s = reader.read_bytes()?;
+        // Firmware signature field is 0x200 bytes; after r and s, the remaining bytes are reserved/padding.
+        reader.skip_bytes::<368>()?;
         Ok(Self { r, s })
     }
 }
--- a/src/certs/snp/signature/ecdsa.rs
+++ b/src/certs/snp/signature/ecdsa.rs
@@ -1,26 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
-#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
 use super::*;
 
-use crate::{
-    parser::{ByteParser, Decoder, Encoder},
-    util::{
-        hexline::HexLine,
-        parser_helper::{ReadExt, WriteExt},
-    },
-};
+use crate::{parser::ByteParser, util::hexline::HexLine};
 
 use std::io::{Read, Result, Write};
 
 #[cfg(feature = "openssl")]
-use crate::certs::snp::{AsLeBytes, FromLe};
+use crate::certs::snp::{AsLeBytes, Certificate, FromLe};
 
 #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
 use std::convert::TryFrom;
 
 #[cfg(feature = "openssl")]
-use openssl::{bn, ecdsa};
+use openssl::{bn, ecdsa, ecdsa::EcdsaSig, sha::Sha384};
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
@@ -130,16 +123,6 @@ impl From<ecdsa::EcdsaSig> for Signature {
 }
 
 #[cfg(feature = "openssl")]
-impl TryFrom<&[u8]> for Signature {
-    type Error = Error;
-
-    #[inline]
-    fn try_from(value: &[u8]) -> Result<Self> {
-        Ok(ecdsa::EcdsaSig::from_der(value)?.into())
-    }
-}
-
-#[cfg(feature = "openssl")]
 impl TryFrom<&Signature> for ecdsa::EcdsaSig {
     type Error = Error;
 
@@ -166,23 +149,57 @@ impl TryFrom<&Signature> for p384::ecdsa::Signature {
             GenericArray::clone_from_slice(&s_big_endian),
         )
         .map_err(|e| {
-            Error::new(
-                ErrorKind::Other,
-                format!("failed to deserialize signature from scalars: {e:?}"),
-            )
+            Error::other(format!(
+                "failed to deserialize signature from scalars: {e:?}"
+            ))
         })
     }
 }
 
 #[cfg(feature = "openssl")]
-impl TryFrom<&Signature> for Vec<u8> {
-    type Error = Error;
+/// Verify ECDSA signature on attestation report using VEK certificate
+pub fn verify_ecdsa_signature(body: &[u8], signature: &[u8], vek: &Certificate) -> Result<()> {
+    let sev_sig = Signature::from_bytes(signature)?;
 
-    #[inline]
-    fn try_from(value: &Signature) -> Result<Self> {
-        Ok(ecdsa::EcdsaSig::try_from(value)?.to_der()?)
+    let sig: EcdsaSig = EcdsaSig::try_from(&sev_sig)?;
+
+    let mut hasher = Sha384::new();
+    hasher.update(body);
+    let base_digest = hasher.finish();
+
+    let ec = vek.public_key()?.ec_key()?;
+    let signed = sig.verify(&base_digest, &ec)?;
+    match signed {
+        true => Ok(()),
+        false => Err(Error::other("VEK does not sign the attestation report")),
     }
 }
+
+#[cfg(feature = "crypto_nossl")]
+/// Verify ECDSA signature on attestation report using VEK certificate
+pub fn verify_ecdsa_signature(body: &[u8], signature: &[u8], vek: &Certificate) -> Result<()> {
+    let sev_sig = Signature::from_bytes(signature)?;
+
+    let sig = p384::ecdsa::Signature::try_from(&sev_sig).map_err(|e| {
+        std::io::Error::other(format!(
+            "failed to generate signature from raw bytes: {e:?}"
+        ))
+    })?;
+
+    use sha2::Digest;
+    let base_digest = sha2::Sha384::new_with_prefix(body);
+    let verifying_key =
+        p384::ecdsa::VerifyingKey::from_sec1_bytes(vek.public_key_sec1()).map_err(|e| {
+            std::io::Error::other(format!(
+                "failed to deserialize public key from sec1 bytes: {e:?}"
+            ))
+        })?;
+    use p384::ecdsa::signature::DigestVerifier;
+    verifying_key.verify_digest(base_digest, &sig).map_err(|e| {
+        std::io::Error::other(format!("VEK does not sign the attestation report: {e:?}"))
+    })
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -262,29 +279,11 @@ mod tests {
         }
 
         #[test]
-        fn test_try_from_bytes() {
-            let r = BigNum::from_dec_str("123").unwrap();
-            let s = BigNum::from_dec_str("456").unwrap();
-            let ecdsa_sig = ecdsa::EcdsaSig::from_private_components(r, s).unwrap();
-            let der = ecdsa_sig.to_der().unwrap();
-            let sig = Signature::try_from(der.as_slice()).unwrap();
-            assert_ne!(sig.r(), &[0u8; 72]);
-            assert_ne!(sig.s(), &[0u8; 72]);
-        }
-
-        #[test]
         fn test_try_into_ecdsa_sig() {
             let sig: Signature = Default::default();
             let ecdsa_sig: ecdsa::EcdsaSig = (&sig).try_into().unwrap();
             assert_eq!(ecdsa_sig.r().to_vec(), vec![]);
             assert_eq!(ecdsa_sig.s().to_vec(), vec![]);
-        }
-
-        #[test]
-        fn test_try_into_vec() {
-            let sig: Signature = Default::default();
-            let der: Vec<u8> = (&sig).try_into().unwrap();
-            assert!(!der.is_empty());
         }
     }
 
--- a/src/certs/snp/signature/mod.rs
+++ b/src/certs/snp/signature/mod.rs
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{
+    convert::TryFrom,
+    fmt::Display,
+    io::{Read, Write},
+};
+
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+use std::io::Error;
+
+use crate::{
+    parser::{Decoder, Encoder},
+    util::parser_helper::{ReadExt, WriteExt},
+};
+
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+use crate::certs::snp::{Certificate, Verifiable};
+
+/// ECDSA algorithm signature
+pub mod ecdsa;
+
+/// Signature algorithms that firmware may use to sign the SEV-SNP attestation report.
+///
+/// The algorithm identifier is encoded in the report body and is therefore derived
+/// from **untrusted bytes**. It MUST NOT be treated as a trust signal on its own.
+/// Authenticity is only established by successfully verifying the report signature.
+///
+/// This enum is intentionally explicit: unknown values are rejected during decoding.
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum SignatureAlgorithm {
+    /// ECDSA with SECP384R1 curve
+    EcdsaSecp384r1 = 1,
+}
+
+/// Creates a [`SignatureAlgorithm`] from the numeric algorithm identifier encoded
+/// in the report body.
+///
+/// # Errors
+///
+/// Returns `InvalidData` if `v` does not correspond to a supported algorithm.
+impl TryFrom<u32> for SignatureAlgorithm {
+    type Error = std::io::Error;
+
+    fn try_from(v: u32) -> Result<Self, Self::Error> {
+        match v {
+            1 => Ok(SignatureAlgorithm::EcdsaSecp384r1),
+            v => Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                format!("Unsupported signature algorithm: {}", v),
+            )),
+        }
+    }
+}
+
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+impl Verifiable for (SignatureAlgorithm, &[u8], &[u8], &Certificate) {
+    type Output = ();
+
+    fn verify(self) -> Result<Self::Output, std::io::Error> {
+        let (algorithm, body, signature, vek) = self;
+        match algorithm {
+            SignatureAlgorithm::EcdsaSecp384r1 => {
+                ecdsa::verify_ecdsa_signature(body, signature, vek)
+            }
+        }
+    }
+}
+
+impl Encoder<()> for SignatureAlgorithm {
+    fn encode(&self, writer: &mut impl Write, _: ()) -> Result<(), std::io::Error> {
+        match self {
+            SignatureAlgorithm::EcdsaSecp384r1 => writer.write_bytes(1u32, ())?,
+        };
+        Ok(())
+    }
+}
+
+impl Decoder<()> for SignatureAlgorithm {
+    fn decode(reader: &mut impl Read, _: ()) -> Result<Self, std::io::Error> {
+        let algo: u32 = reader.read_bytes()?;
+        Self::try_from(algo)
+    }
+}
+
+impl Display for SignatureAlgorithm {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SignatureAlgorithm::EcdsaSecp384r1 => write!(f, "ECDSA with SECP384R1"),
+        }
+    }
+}
--- a/src/firmware/guest/types/snp.rs
+++ b/src/firmware/guest/types/snp.rs
@@ -1,16 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
-use crate::certs::snp::{Certificate, Chain, Verifiable};
+use crate::certs::snp::{ecdsa::Signature, Certificate, Chain, Verifiable};
 
 use crate::{
-    certs::snp::ecdsa::Signature,
-    error::AttestationReportError,
     firmware::host::TcbVersion,
     parser::{ByteParser, Decoder, Encoder},
     util::{
         hexline::HexLine,
-        parser_helper::{ReadExt, WriteExt},
+        parser_helper::{validate_reserved, ReadExt, WriteExt},
     },
     Generation,
 };
@@ -21,13 +19,10 @@ use serde::{Deserialize, Serialize};
 use serde_big_array::BigArray;
 
 use std::{
+    convert::TryFrom,
     fmt::Display,
-    io::{Cursor, Read, Write},
-    ops::Range,
+    io::{Read, Write},
 };
-
-#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
-use std::convert::TryFrom;
 
 #[cfg(feature = "openssl")]
 use std::io::Error;
@@ -36,11 +31,6 @@ use bitfield::bitfield;
 
 #[cfg(feature = "openssl")]
 use openssl::{ecdsa::EcdsaSig, sha::Sha384};
-
-const ATT_REP_FW_LEN: usize = 1184;
-const CHIP_ID_RANGE: Range<usize> = 0x1A0..0x1E0;
-const CPUID_FAMILY_ID_BYTES: usize = 0x188;
-const CPUID_MODEL_ID_BYTES: usize = 0x189;
 
 /// Structure of required data for fetching the derived key.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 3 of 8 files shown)


DGonzalezVillal:
@tylerfanelli @cschramm @hyperfinitism

npmccallum (commented):
## Review: `ReportBody` should carry fully-parsed types, not raw byte slices

The overall design (two-phase: `Report` for signature verification over raw bytes, then `ReportBody` for typed access) is solid. But the current `ReportBody` is a half-measure — it stores `&[u8]` slices and then has getters that do `u32::from_le_bytes()` / `.try_into().unwrap()`. Those `unwrap()`s are *technically* safe (lengths are guaranteed by `from_bytes`), but the conversions should happen at parse time so errors propagate through the `Result` returned by `from_bytes`. Then the getters become redundant and can be deleted.

### 1. Parse TCB fields fully at parse time — derive `Generation` internally

This is the big one. The current code stores TCB fields as `&[u8]` and forces callers to pass `Generation` into getter methods like `current_tcb(&self, generation: Generation) -> TcbVersion`. But the parser already has everything it needs to derive `Generation` internally:

- For V2: use the `chip_id_is_turin_like` heuristic (bytes `0x1A0..0x1E0`)
- For V3+: use `cpuid_fam_id`/`cpuid_mod_id` with `Generation::identify_cpu`

This is exactly what `AttestationReport::Decoder` already does — `ReportBody::from_bytes` should do the same. Store as `TcbVersion` directly, not `&[u8; 8]`.

Deferring TCB parsing to getters forces a third level of parsing on the caller and defeats the purpose of a typed report body.

### 2. Expose `ReportVariant` as a field

The parser already computes `ReportVariant` from `version_num` to determine which optional fields are present. Expose it as `pub variant: ReportVariant` so callers know the version semantics without re-deriving from the raw version number.

### 3. Scalar fields → final types at parse time

All of these should be parsed in `from_bytes`, not in getters:

- `version` → `u32`
- `guest_svn` → `u32`
- `vmpl` → `u32`
- `sig_algo` → `u32`
- `policy` → `GuestPolicy`
- `plat_info` → `PlatformInfo`
- `key_info` → `KeyInfo`
- `current` / `committed` → `Version`
- `launch_mit_vector` / `current_mit_vector` → `Option<u64>`
- `cpuid_fam_id` / `cpuid_mod_id` / `cpuid_step` → `Option<u8>`
- All TCB fields → `TcbVersion` (per point 1)

### 4. Fixed-size byte arrays → `&'a [u8; N]` via `TryFrom`

For reference types where zero-copy makes sense, use `<&[u8; N]>::try_from(slice)?` which is fallible and propagates errors cleanly:

- `family_id: &'a [u8; 16]`
- `image_id: &'a [u8; 16]`
- `report_data: &'a [u8; 64]`
- `measurement: &'a [u8; 48]`
- `host_data: &'a [u8; 32]`
- `id_key_digest: &'a [u8; 48]`
- `author_key_digest: &'a [u8; 48]`
- `report_id: &'a [u8; 32]`
- `report_id_ma: &'a [u8; 32]`
- `chip_id: &'a [u8; 64]`

### 5. Delete all getters

With the above changes, every field carries its final type. The getters become redundant — callers just access the public fields directly.

### Summary

The principle: **the parser should do all the parsing.** `ReportBody` fields should be the final typed values, with all conversion errors propagated through `from_bytes`. Zero-copy `&[u8; N]` references where it makes sense for large byte arrays, owned scalars/enums for everything else. No post-parse interpretation required by the caller.

DGonzalezVillal:
@tylerfanelli Could you PTAL, to this implementation. I have implemented what Nathaniel proposed and would like to get thoughts. 

@hyperfinitism @cschramm @npmccallum 

Could you also ptal.

DGonzalezVillal:
Updated PR description at the top

DGonzalezVillal:
> Overall, this looks like a clean and well-structured design. The separation of concerns around report parsing is clearer than before, and the direction makes sense to me. LGTM from an architectural standpoint.
> 
> One concern: with the current approach, it seems that V2 attestation reports from pre-Turin generations would not be parsable, since the TcbVersion variant is inferred heuristically. Would it make sense to also provide an explicit API where the caller can supply the `Generation` (or directly the `TcbVersion` variant) when needed?
> 
> Of course, it is also reasonable to treat older FW versions producing V2 reports as out of scope and not guaranteed to work.

Hmm, why would those not be parsable?

The current logic is that if it is a V2 report, we determine the generation from the chip ID based on its layout. There are currently two TCB formats:

- Pre-Turin (Milan & Genoa)

- Turin-forward (Turin & Venice)

We assign the report a generation value of either pre-Turin or Turin-forward strictly for the purpose of parsing the TCB. The generation is not otherwise used — it only determines which TCB layout should be applied.

The decoder (a custom parser implementing `Read`) requires the generation in order to correctly parse the TCB:

```rust
fn decode(reader: &mut impl Read, generation: Generation) -> Result<Self, std::io::Error> {
    match generation {
        Generation::Milan | Generation::Genoa => {
            Ok(TcbVersion::from_legacy_bytes(&reader.read_bytes()?))
        }
        Generation::Turin | Generation::Venice => {
            Ok(TcbVersion::from_turin_bytes(&reader.read_bytes()?))
        }
        _ => Err(std::io::Error::new(
            std::io::ErrorKind::Unsupported,
            "Unsupported generation for TCB parsing",
        )),
    }
}
```

To use this decoder, a generation must always be supplied when parsing the TCB, since it determines which TCB format the raw bytes should be interpreted as.

The only way the V2 report could not be parsed is if the guest has Mask Chip ID enabled, which would fill that value with all 0's which would then make it unparsable with current logic, but in that case we return the appropriate error.

hyperfinitism:
My apologies, I was confusing this with the heuristics around KDS. As you mentioned, parsing only fails when the MASK_CHIP_ID is enabled (e.g., AWS EC2) and the Report Version is 2. For such reports, I was considering a function where the user explicitly provides the Generation for parsing. However, since this is an extremely rare case, returning error might be sufficient.
(Note: The current AWS EC2 instances with SNP issue V5 reports.)

npmccallum (commented):
## Overall Assessment

This is a well-designed PR that implements the two-phase report parsing flow discussed in #373. The architecture cleanly separates framing (`Report`) from typed access (`ReportBody`), with signature verification as the gate between them. It addresses earlier review feedback (fully parsed types at parse time, no getters, `SignatureAlgorithm` as single dispatch point). The code is substantially better than the old `AttestationReport`.

## Issues

### 1. Double doc-comment marker
`src/certs/snp/mod.rs`:
```rust
/// /// Attestation report signature parsing and verification helpers.
```
Extra `///` in the doc comment.

### 2. `V3 = 3 | 4` is not valid Rust for discriminants — and causes a downstream bug

`ReportVariant` uses `#[repr(u32)]` but `V3 = 3 | 4` evaluates to `V3 = 7` (bitwise OR). The `TryFrom<u32>` impl correctly maps 3 and 4 → V3, but the discriminant value `7` is misleading and won't round-trip correctly if anyone does `variant as u32`. Should be `V3 = 3` with a comment explaining that firmware version 4 also maps here.

### 3. `version as u32` casts in `ReportBody::from_bytes` are broken (critical bug)

The cast `version as u32` (where `version: ReportVariant`) will produce `7` for V3 due to issue #2. The comparisons `version as u32 >= 3` and `version as u32 >= 5` happen to still work for the CPUID branch (7 >= 3 is true), but **`version as u32 >= 5` is also true for V3** (since 7 >= 5), so a V3 report would incorrectly take the V5 branch for mitigation vectors — parsing `launch_mit_vector` and `current_mit_vector` from what should be reserved bytes, and only validating `0x208..0x2A0` instead of `0x1F8..0x2A0`.

The fix: use pattern matching or `version >= ReportVariant::V3` / `version >= ReportVariant::V5` comparisons (which work correctly via derived `Ord`) instead of `as u32` casts.

### 4. `Encoder` removed for `ReportBody` but `ByteParser` still implemented

`ByteParser<()> for ReportBody<'_>` is implemented with `type Bytes = [u8; ATT_REP_FW_LEN]`, but there's no `Encoder` impl anymore (the old `AttestationReport` had one). If `ByteParser` requires `Encoder`, this won't compile. If serialization is intentionally dropped, `ByteParser` should also be removed.

### 5. `Report` derives `Serialize`/`Deserialize` with borrowed `&'a [u8]`

Serde for `&'a [u8]` has surprising behavior — it doesn't round-trip through most formats (JSON will serialize as an array of numbers but won't deserialize back to `&[u8]`). Since `Report` is meant to be a transient view, consider removing the serde derives entirely.

### 6. `Report::from_bytes` vs `ReportBody::from_bytes` length check inconsistency

`Report::from_bytes` requires `report.len() != REPORT_LEN` (exact), while `ReportBody::from_bytes` requires `body.len() < BODY_LEN` (minimum). These should be consistent — both should use exact checks since these are fixed-format structures.

### 7. Missing reserved field validation for legacy TCB

`from_legacy_bytes` validates `bytes[3..4]` but the legacy TCB layout also has reserved bytes at positions 2 and 4-5 per the spec. Similarly `from_turin_bytes` only validates `bytes[4..5]` — are bytes 5-6 also reserved?

### 8. `GuestPolicy` `Decoder<()>` still exists without validation

The comment says "No checking in case policy is being parsed outside attestation report (e.g. id-block)" — but this means there are two decode paths and callers could accidentally use the wrong one. Consider at minimum adding a doc comment on the `Decoder<()>` impl explaining when to use which.

### 9. `PlatformInfo` validation inconsistency with `GuestPolicy`

Unlike `GuestPolicy` which has a separate `Decoder<ReportVariant>` for the strict path, `PlatformInfo::Decoder<()>` always validates reserved bits. This is inconsistent — if `PlatformInfo` can appear outside reports (like `GuestPolicy` can), the unconditional validation could break those paths.

### 10. `validate_reserved` error message doesn't include offset

The error "reserved bytes are non-zero" gives no indication of *which* reserved field failed. Including the byte offset or a caller-provided context string would make debugging much easier.

## Minor / Style

- Several test comments have "alorithm" typo (should be "algorithm")
- The `unsafe` block in `test_report_from_bytes_splits_body_and_signature` uses `offset_from` — could use arithmetic on pointer offsets without unsafe
- `Signature::decode` now skips 368 padding bytes — worth a `debug_assert!` that the skipped bytes are zero, or a comment explaining why they're not validated

DGonzalezVillal:
@npmccallum I've addressed all your comments.

I’d like your thoughts on my approach for parsing the report fields PlatformInfo, KeyInfo, and GuestPolicy.

In this issue: https://github.com/virtee/sev/issues/330

The user suggests that we should parse GuestPolicy and PlatformInfo based on the attestation report version. However, what I found is that—at least for GuestPolicy—the ABI allows policy bits to evolve without requiring a change to the attestation report version. This means that the report version alone is not a reliable indicator of which policy bits are valid.

Given that, I decided to take a different approach: use the firmware version reported in the attestation report and compare it against when specific policy bits were introduced in the ABI.

For example, from the changelog:

September 2023 — Rev 1.55 Updates and Additions:
• Added feature capability reporting
• Added ciphertext hiding feature
• Added AES-256 XTS guest policy
• Added CXL guest policy
• Added RAPL disable guest policy
• Added ECC guest policy
• Added SNP disable on SNP shutdown feature

This suggests that ABI 1.55 is when those policy bits became defined. So, we can use the firmware version to determine whether those bits should be interpreted or treated as unknown.

You can also see that these changes did not trigger a bump in the attestation report version, which reinforces that report version is not sufficient for determining field semantics.

I’m applying the same approach to PlatformInfo and KeyInfo as well.

DGonzalezVillal:
@hyperfinitism Could you take a look?

hyperfinitism:
LGTM. Using firmware version seems more robust. If there is one concern, it is that the mapping between the report structure, kernel version, and report version is not strictly defined in the FW ABI spec.

cschramm:
@DGonzalezVillal: Sorry for the late check-in. Unfortunately, I fail to see how this addresses #343.

The parser is basically `ReportBody::from_bytes` now and it breaks for newer versions not only because it assumes (previously) reserved areas to be zero, but also because it now checks the version explicitly.

My distributed clients need a parser for attestation reports that does not just break for newer versions, but tries to apply the structure from the last known one. Things the clients are interested in from `AttestationReport` / `ReportBody` include e.g. `plat_info`, `policy` and `key_info` bits, `vmpl`, `committed`, `committed_tcb`. When servers get a firmware upgrade that bumps the report version, all that is unreachable for existing distributed clients that have not received an update yet, as `ReportBody::from_bytes` will fail.

src/certs/snp/signature/ecdsa.rs
```diff
@@ -68,6 +61,8 @@ impl Decoder<()> for Signature {
     fn decode(reader: &mut impl Read, _: ()) -> Result<Self> {
         let r = reader.read_bytes()?;
         let s = reader.read_bytes()?;
+        // Firmware signature field is 0x200 bytes; after r and s, the remaining bytes are reserved/padding.
+        reader.skip_bytes::<368>()?;
```
larrydewey, line 65: Not within the scope of this issue, but we should probably rename this :)

src/certs/snp/signature/ecdsa.rs
```diff
@@ -166,23 +151,57 @@ impl TryFrom<&Signature> for p384::ecdsa::Signature {
...
         })
     }
 }
 
 #[cfg(feature = "openssl")]
-impl TryFrom<&Signature> for Vec<u8> {
-    type Error = Error;
+/// Verify ECDSA signature on attestation report using VEK certificate
+pub fn verify_signature(body: &[u8], signature: &[u8], vek: &Certificate) -> Result<()> {
+    let sev_sig = Signature::from_bytes(signature)?;
```
larrydewey, line 164: Needs to be fallable, as the bytes may not work properly

larrydewey: Should we be using verifiable here, instead of using a custom signature?

src/certs/snp/signature/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+impl SignatureAlgorithm {
+    /// Verify the report signature over `body` using `vek` and this algorithm.
+    ///
+    /// This validates the signature bytes against the signed body bytes; it does
+    /// not parse or validate any fields inside the body.
+    #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+    pub fn verify(
```
larrydewey, line 63: Probably should also be verifiable.

src/firmware/guest/types/snp.rs
```diff
@@ -230,290 +308,437 @@ impl ByteParser<()> for ReportVariant {
...
+    /// verified [`ReportBody`].
+    pub fn from_bytes(report: &'a [u8]) -> std::io::Result<Self> {
+        if report.len() != Self::REPORT_LEN {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "Bad report length",
+            ));
+        };
+
+        let algorithm = SignatureAlgorithm::decode(&mut &report[0x34..0x38], ())?;
```
larrydewey, line 373: Maybe use a const-range here?

src/firmware/guest/types/snp.rs
```diff
@@ -230,290 +308,437 @@ impl ByteParser<()> for ReportVariant {
...
+            algorithm,
+            body: &report[..Self::BODY_LEN],
+            signature: &report[Self::SIG_OFF..Self::SIG_OFF + Self::SIG_LEN],
+        })
+    }
+}
+
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+impl Verifiable for (&Certificate, &Report<'_>) {
+    type Output = ();
```
larrydewey, line 385: Not within the scope of this, but maybe use boolean for the type output?

src/firmware/guest/types/snp.rs
```diff
@@ -230,290 +308,437 @@ impl ByteParser<()> for ReportVariant {
...
+#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
+/// Verifies `report` with `vek` and returns a parsed [`ReportBody`].
+impl<'a> TryFrom<(&Report<'a>, &Certificate)> for ReportBody<'a> {
+    type Error = std::io::Error;
+
+    fn try_from((report, vek): (&Report<'a>, &Certificate)) -> Result<Self, Self::Error> {
+        (vek, report).verify()?;
+        ReportBody::from_bytes(report.body)
+    }
+}
```
larrydewey, line 416: I almost feel like it would make more sense to implement verifiable for this:


```suggestion
#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
/// Verifies `report` with `vek` and returns a parsed [`ReportBody`].
impl<'a> Verifiable for (&Report<'a>, &Certificate) {
    type Output = ReportBody<'a>;
    
    fn verify(self) -> Result<Self::Output, std::io::Error> {
        // Logic here...
    }
}
```

src/firmware/guest/types/snp.rs
```diff
@@ -230,290 +308,437 @@ impl ByteParser<()> for ReportVariant {
...
+    /// Guest policy governing hypervisor restrictions.
     pub policy: GuestPolicy,
-    /// The family ID provided at launch.
-    pub family_id: [u8; 16],
-    /// The image ID provided at launch.
-    pub image_id: [u8; 16],
-    /// The request VMPL for the attestation report.
+
+    /// Family ID provided at launch (128 bits).
+    pub family_id: &'a [u8; 16],
```
larrydewey, line 460: Not requried, but you might do:


```suggestion
    pub family_id: &'a [u8],
```

src/firmware/host/types/snp.rs
```diff
@@ -562,14 +562,15 @@ impl ByteParser<Generation> for TcbVersion {
 }
 
 impl TcbVersion {
-    pub(crate) fn from_legacy_bytes(bytes: &[u8; 8]) -> Self {
-        Self {
+    pub(crate) fn from_legacy_bytes(bytes: &[u8; 8]) -> Result<Self, std::io::Error> {
```
larrydewey, line 565: Maybe call it try_from_legacy_bytes? :)

larrydewey: We probably should do TryFrom<&[u8]> instead...

larrydewey (changes requested):

src/firmware/guest/types/snp.rs
DGonzalezVillal: So wouldn't it be better to have it fixed size, so that we know the data is fitting as expected in the spec?

src/certs/snp/signature/ecdsa.rs
DGonzalezVillal: skip_bytes is confusing, but for future reference skip bytes does check that all the bytes that are being skipped are 0

src/certs/snp/signature/ecdsa.rs
DGonzalezVillal: So the from_bytes function is a naming convention mistake from the ByteParser trait were we named the function from_bytes instead of try_from_bytes. But it's not related to the rust internals From and TryFrom. I can address this on a separate PR where we fix naming conventions, also SkipBytes.

DGonzalezVillal: > Should we be using verifiable here, instead of using a custom signature?

So I looked into it, and I think for this design the `Verifiable` trait does not really make sense at the lower-level signature verification helper.

Right now, we implement `Verifiable` for the certificate/report pair:

```rust
#[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
impl<'a> Verifiable for (&Certificate, &Report<'a>) {
    type Output = ReportBody<'a>;

    fn verify(self) -> Result<Self::Output, std::io::Error> {
        let (vek, report) = self;

        let algo = report.algorithm;

        algo.verify(report.body, report.signature, vek)?;

        ReportBody::from_bytes(report.body)
    }
}
```
This is the level where the trait makes the most sense: the tuple contains the full verification context needed to authenticate the report and then return the parsed ReportBody.

Inside that implementation, we use the algorithm from the report to select the appropriate verification routine:

```rust
impl SignatureAlgorithm {
    /// Verify the report signature over `body` using `vek` and this algorithm.
    ///
    /// This validates the signature bytes against the signed body bytes; it does
    /// not parse or validate any fields inside the body.
    #[cfg(any(feature = "openssl", feature = "crypto_nossl"))]
    pub fn verify(
        &self,
        body: &[u8],
        signature: &[u8],
        vek: &Certificate,
    ) -> Result<(), std::io::Error> {
        match self {
            SignatureAlgorithm::EcdsaSecp384r1 => {
                ecdsa::verify_ecdsa_signature(body, signature, vek)
            }
        }
    }
}
```

That method is more of an algorithm dispatch point. It already receives the body, signature, and VEK that were collected by the higher-level Verifiable implementation.

From there, we call the backend-specific helper, which converts the raw signature bytes into the backend’s expected ECDSA signature type and performs the actual cryptographic verification.

So I think Verifiable makes sense for the report/certificate verification context, but not really for the raw signature helper. To implement the trait at that level, we would end up with something like:

```rust
impl Verifiable for (&[u8], &[u8], &Certificate)
```

That feels more misleading than helpful because the raw byte tuple is not really a domain object that “self-verifies”; it is just the input to a backend-specific verification function.

Thoughts?

src/certs/snp/signature/mod.rs
DGonzalezVillal: Same thought as:
https://github.com/virtee/sev/pull/378#discussion_r3047895371

src/certs/snp/signature/ecdsa.rs
larrydewey: I am fine with that justification. I agree, we can scope the refactor of the Trait to an individual PR.

src/firmware/guest/types/snp.rs
DGonzalezVillal: So I implemented it, but after thinking about it more, I think `TryFrom` makes more sense for the design I am trying to implement.

The problem with having `Verifiable` output `ReportBody` is that there are two possible failure points:

1. Signature verification can fail.
2. Parsing `ReportBody` can fail.

This is the design flaw I am trying to get away from in this PR. I want the body and signature to be verifiable regardless of whether we can parse `ReportBody`.

`Verifiable` for `(&Certificate, &Report)` should only tell us whether the report is valid. Then `TryFrom` can use this verification step to confirm that the report is valid before attempting to parse the body.

This gives users two options:

- Use `Verifiable` if they only want to verify the report signature.
- Use `TryFrom` if they want to verify the report and then parse it into `ReportBody`.

src/firmware/host/types/snp.rs
DGonzalezVillal: I changed the naming convention.

There is one issue with implementing `TryFrom` for `TcbVersion` directly: we cannot communicate which layout of the TCB structure we want to parse the bytes into.

We could do something like this:

```rust
TryFrom<(&[u8], TCBVariant)> for TCBVersion
```

and create a `TcbVariant` enum, but I think that would be essentially equivalent to what we already have.

Also, the function works well with a sized parameter because, in the decoder, we are doing this:

```rust
mpl Decoder<Generation> for TcbVersion {
    fn decode(reader: &mut impl Read, generation: Generation) -> Result<Self, std::io::Error> {
        match generation {
            Generation::Milan | Generation::Genoa => {
                TcbVersion::try_from_legacy_bytes(&reader.read_bytes()?)
            }
            Generation::Turin | Generation::Venice => {
                TcbVersion::try_from_turin_bytes(&reader.read_bytes()?)
            }
            _ => Err(std::io::Error::new(
                std::io::ErrorKind::Unsupported,
                "Unsupported Processor Generation for TCB parsing",
            )),
        }
    }
```

We are using the read_bytes helper you designed, and it needs to know the number of bytes it is reading. That is why we are using the sized &[u8; 8] instead of &[u8].

PR #392, opened 2026-05-08
add EPYC-Turin vCPU type support

rouming:
PR adds EpycTurin, EpycTurinV1, EpycTurinV2 variants to CpuType so callers:(notably snpguest's `generate measurement --vcpu-type EPYC-Turin`) can compute launch digests for guests booted with QEMU's `EPYC-Turin` CPU model. Without this, snpguest fails with:
```    
        ERROR: Invalid VCPU type value provided: EPYC-Turin
```

The CPUID family / model / stepping values used by sig() were taken verbatim from QEMU's CPU definition at target/i386/cpu.c (entry .name = "EPYC-Turin").

Also fix some warnings when the lib is built with snp feature support only.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-#[cfg(feature = "openssl")]
+#[cfg(all(feature = "openssl", feature = "sev"))]
 use openssl::error::ErrorStack;
 use std::{
     array::TryFromSliceError,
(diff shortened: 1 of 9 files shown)


tylerfanelli:
cc/ @DGonzalezVillal

DGonzalezVillal (approved):
The changes lgtm, @tylerfanelli do you mind if the introducing the warning fixes in this pr even if they're not directly related to the Turin Vcpu?

tylerfanelli:
It's fine with me, as it's a relatively small change.

tylerfanelli (approved):

rouming:
@tylerfanelli @DGonzalezVillal guys, do you plan to release the sev lib with this change and update snpguest accordingly? Or you have a strict schedule for that? I'm asking because I depend on both in my other repo (the attest tool), I need turin support and don't want to depend on my own branches.

DGonzalezVillal:
@rouming I'm planning a release soon, I want to cut a release with all the new content before https://github.com/virtee/sev/pull/378 gets merged since that will be a major breaking change

PR #396, opened 2026-09-01
feat(firmware) add venice TCB variant to TcbVersion

DGonzalezVillal:
Venice does not encode bootloader or microcode in TCB, so those fields are now Option and encode/decode independently of Turin.


amd-kkin:
LGTM, checked against https://docs.amd.com/v/u/en-US/56860_PUB_SEV_SNP, section 2.3

tylerfanelli (approved):

